PR #8, opened 2015-01-15
fixed #7: on_{state} can not work when state is not defined as string

### LINKED ISSUE #7: on_{state} can not work when state is not defined as string
Can't work
image

Work
image

ExpandOcean:
fixed #7 , and modify the test case


tyarkoni:
Thanks!

PR #10, opened 2015-02-03
NEW added optional 'ignore' flag to States

bvdeenen:
adding the flag causes the states to ignore triggers that are not defined.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -13,7 +13,7 @@ def listify(obj):
 
 class State(object):
 
-    def __init__(self, name, on_enter=None, on_exit=None):
+    def __init__(self, name, on_enter=None, on_exit=None, ignore=False):
         """
         Args:
             name (string): The name of the state
@@ -23,8 +23,12 @@ class State(object):
             on_exit (string, list): Optional callable(s) to trigger when a
                 state is exited. Can be either a string providing the name of a
                 callable, or a list of strings.
+            ignore (Boolean): Optional flag to indicate if unhandled triggers
+                should raise an exception
+
         """
         self.name = name
+        self.ignore = ignore
         self.on_enter = listify(on_enter) if on_enter else []
         self.on_exit = listify(on_exit) if on_exit else []
 
@@ -184,9 +188,10 @@ class Event(object):
         """
         state_name = self.machine.current_state.name
         if state_name not in self.transitions:
-            raise MachineError(
-                "Can't trigger event %s from state %s!" % (self.name,
-                                                           state_name))
+            if not self.machine.current_state.ignore :
+                raise MachineError(
+                    "Can't trigger event %s from state %s!" % (self.name,
+                                                               state_name))
         event = EventData(self.machine.current_state, self,
                           self.machine, self.machine.model, *args, **kwargs)
         for t in self.transitions[state_name]:


tyarkoni:
Thanks for the PR! While this is a simple enough patch, I'm not sure the State initializer is the best place to handle something like this--I'd like to keep the API minimalistic, and adding this kind of thing could easily snowball into special handlers for all kinds of other potential exceptions. I'm also not crazy in general about the idea of preemptively suppressing valid exceptions, as opposed to handling them reactivity. Would one of these alternative patterns work okay for you:

* Wrap a try/catch block around trigger calls in your app that could fail depending on the current state (which would require no modification to the transitions codebase);
* Replace the generic MachineException thrown here with something more specific (e.g., InvalidInitialStateException), which would allow you to catch just this exception at the top level of your application and check against a list of states that should be ignored;
* Use more specific triggers. I'm assuming the reason this particular issue arises is that you have a trigger that's supposed to work for multiple states, just not all of them. Being more specific about the list of valid initial conditions in each transition may make your code a bit more verbose, but is probably a safer way to go in general than adding broader triggers that then have to be ignored for some states;
* Make the transition conditional on some other check passing. For example, let's say you have a transition called advance() that moves from any state into state C, and you want to prevent the transition only if the current state is A. Then you could do something like:

```python
machine.add_transition('advance', '*', 'C', conditions='valid_source')

class MyModel(object):
    ...
    def valid_source(self, event_data):
        return False if event_data.state.name == 'A' else True
```

Of the above, I would personally probably go with the third, then the fourth, and only resort to exception handling if there was some reason why the other two approaches aren't practical. But let me know if none of these makes sense for your particular use case and maybe we can figure something else out. I'll close this for now, but will re-open if needed.

bvdeenen:
Hi

First of all, I have really cleaned up our code using your transitions library. Thanks!

I disagree with you though :-)
In FSM's used in communications, it's very common, to create an event in the future, like 'timeout', which in case the FSM is still in the state 'waiting_for_answer' should trigger a state transition to the state 'offline'. I don't care about the 'timeout' trigger if I'm already in the state 'auth_connected'. I just want it discarded. When states and triggers are fairly complex, I really don't want to define all triggers for every source, it complicates the FSM, and adds nothing.

To me, this kind of behavior is the default of an FSM. Unspecified events should be ignored.

Since we disagree on this, I've just added the single file to our project (with my changes). 

Again, thanks for this well thought out piece of code.

tyarkoni:
Hmmm... okay, you've almost swayed me. But I'm still wary of feature creep for the reason mentioned earlier. So let me reopen this and let it sit here, and if one or two other people indicate that they'd like to have this functionality as well, we'll add it in. My reservation is mainly driven by the fact that I haven't seen other simple FSM implementations provide this functionality, which leads me to think that it's not a common need. But if I'm wrong about that, I'll happily add it. :)

bvdeenen:
Hi 

I'm fairly sure http://machina-js.org/ ignores unspecified triggers.

tyarkoni:
Hi @bvdeenen,

I received an email from another user who also would like to see this exact feature implemented, so as promised, let's go ahead and merge it. But would you mind making the following two changes first:

1. Write a test for the new functionality.
2. Rename the ignore argument to something slightly longer and more descriptive (maybe ignore_invalid_triggers).

If you don't have time, just let me know and I'll do it.

Also, given that we're adding this, I wonder if it would make sense to also add a global switch (at the Machine level) that changes the default behavior. One concern the emailer raised is that he has a lot of different States, and having to set ignore=True explicitly in each one could get unwieldy. If you feel like it, maybe add a global ignore_invalid_triggers to the Machine that changes the default behavior in the State initializer, provided it doesn't complicate things too much.

bvdeenen:
Hi Tal

I'm not working with that piece of code at the moment (It's operational
in our product, with my changes, but we just copied your code). So yes,
I don't really have time to dig it up, so if you could do it, that would
be awesome.

I think a global trigger would be great, because that's the way I think
FSM's should work :-)

If you don't have time, I will make time in after two weeks (our sprint
period), because I think it's important that we work together on making
opensource projects better. This is going to be a key theme of the
startup I'm working at.

Thanks for this great project

Bart
--
  Bart van Deenen bart@infoposter.nl



On Tue, Apr 14, 2015, at 22:24, Tal Yarkoni wrote:
> Hi @bvdeenen[1],


> I received an email from another user who also would like to see this
> exact feature implemented, so as promised, let's go ahead and merge
> it. But would you mind making the following two changes first:


>  1. Write a test for the new functionality.
>  2. Rename the ignore argument to something slightly longer and more
>     descriptive (maybe ignore_invalid_triggers). If you don't have
>     time, just let me know and I'll do it.


> Also, given that we're adding this, I wonder if it would make sense to
> also add a global switch (at the Machine level) that changes the
> default behavior. One concern the emailer raised is that he has a lot
> of different States, and having to set ignore=True explicitly in each
> one could get unwieldy. If you feel like it, maybe add a global
> ignore_invalid_triggers to the Machine that changes the default
> behavior in the State initializer, provided it doesn't complicate
> things too much.


> — Reply to this email directly or view it on GitHub[2].

>



Links:

  1. https://github.com/bvdeenen
  2. https://github.com/tyarkoni/transitions/pull/10#issuecomment-93049599

tyarkoni:
Sounds good. I'll add the test and change the argument name now, but probably won't have the time to add the global flag for a few days.

tyarkoni:
Support for global suppression of the invalid trigger exception added in bd1022b7bff2141c34f4a2107e12a2133a56dbe.

PR #11, opened 2015-03-06
fixed #6: on_enter_{state} does not excute automatically

### LINKED ISSUE #6: on_enter_{state} does not excute automatically
I found on_enter_{state} excute only if I set the on_enter=xxx in defination of state. I think the code below is natural, but it can not work. Do you think it should be supported?

image

mayowa:


transitions/core.py
```diff
@@ -282,11 +282,22 @@ def add_states(self, states, on_enter=None, on_exit=None):
...
+            if a model is provided and on_enter or on_exit is None, the model will be scanned for
+            methods matching the patern on_enter_<state>, on_exit_<state>
         """
+
         states = listify(states)
         for state in states:
             if isinstance(state, str):
-                state = State(state, on_enter=on_enter, on_exit=on_exit)
+                enter_fn = on_enter
+                if enter_fn is None and self != self.model and hasattr(self.model, 'on_enter_'+state):
```
tyarkoni: This works nicely, but one potential limitation is that if the user passes in callbacks as arguments, any pre-defined method will be ignored even if it exists. Perhaps we should treat the pre-defined method as just another callback, and add it to the on_enter and on_exit lists if it exists? Let me know if you think this makes sense. If not, I'll merge as-is.

mayowa: Given the way the library is structured (i.e states can have multiple callbacks), that makes sense.

PR #12, opened 2015-05-13
added property initial

dcere:
Added property initial to see the machine's initial state


tyarkoni:
Thanks for the PR! Happy to merge this, but could you also add another test condition that checks to make sure that if no value is explicitly passed for initial in the Machine initializer, the property initial will be set to 'initial' (the default)? Thanks!

tyarkoni:
Thanks!

dcere:
Thanks to you!

Issue #13, opened 2015-05-21
a generic callback


xdmiodz:
Hi!

Is there a way to specify a generic callback which will be called whenever the machine changes its state?

xdmiodz:
never mind, I can use a decorator instead

tyarkoni:
I like the idea in principle, and would be happy to consider a PR that adds this relatively unobtrusively (I don't have time to work on it myself right now).

tyarkoni:
Closed in #17

PR #17, opened 2015-06-15
Add support for generic callbacks in the Machine

xdmiodz:
The callbacks are called every time a transition happens and have the same call signature as normal callbacks

This patch address #13


xdmiodz:
@tyarkoni please let me know if want me to update README.md too

tyarkoni:
Looks good; thanks for the PR!

PR #23, opened 2015-07-29
make send_event also pass EventData through to conditions

acjc:
Hey,

I've recently found that it would be useful if condition functions also had access to EventData (or args/kwargs) so that data can be passed through when triggering a transition to help determine whether the transition should succeed or not.

Here's a PR implementing this feature. Would appreciate it if you could take a look, thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build/
 dist
 dist/*
 *.egg-info
+*.iml
+.idea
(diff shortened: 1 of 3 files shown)


tyarkoni:
Thanks for the PR, @acjc! I think this an excellent idea, and I'm looking forward to merging it, but it may be a couple of days before I get to it, as I'm swamped with other stuff.

acjc:
No problem, thanks for getting back to me :)

tyarkoni:
Looks great, thanks!

acjc:
Thanks for merging, would you mind doing a new pip release?

tyarkoni:
Yep, done!

PR #26, opened 2015-07-31
Replaced State_a and so on into solid and so on

nokfi:
The tutorial contained a small bug in that the say_goodbye callback was (in the text) associated to State_a, but later executed when the sublimate trigger was called on the state `solid'. I replaced the names State_a and so on into the states solid and so on to line up this example of callback with the rest of the tutorial


tyarkoni:
Great, thanks!

PR #27, opened 2015-07-31
Changed a further occurence of State A into solid, and so on

nokfi:
In my previous pr I missed another occurence of State A.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ class Matter(object):
     
 lump = Matter()
 
-# Same states as above, but now we give StateA an exit callback
+# Same states as above, but now we give state solid an exit callback
 states = [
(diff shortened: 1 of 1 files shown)


tyarkoni:
Looks good, except there are two places where you now have a dictionary with duplicate keys. E.g.,

```python
states = [
    State(name='solid'),
    'liquid', 
    { 'name': 'gas', 'name': 'plasma'}
    ]
```

This won't work; the plasma state needs to be a separate element in the list. You could do [ ... {'name': 'gas'}, {'name': 'plasma'} ... ], which would work fine. If you can change that, I'm happy to merge this as well. Thanks!

tyarkoni:
Thanks!

PR #33, opened 2015-08-31
Fix pickling / unpickling

njouanin:


tyarkoni:
Hmm... the travis build appears to be failing. Looks like it's the test_pickle test. Can you look into it? Happy to merge once it passes. Thanks!

njouanin:
Tests are passing on my computer with python 3.4.
I'll check why tests are failing on travis.

njouanin:
pickle/unpickle doesn't work on python < 3.4 because the pickle protocol used by this versions doesn't support bound methods in objects. Only protocol 4 supports this (see PEP 3154 and http://bugs.python.org/issue9276)
transitions makes a big uses of bounded methods and making pickle works for python < 3.3 would require a lot of changes. What do you think of that ?

njouanin:
pickle/unpickle of Machine works on Python 3.4 with built-in pickle module. For Python < 3.4, a workaround is to use dill module as shown in unit test.
The fix in transitions/core.py is mandatory for making pickle work with all python versions.

tyarkoni:
Looks great, thanks for the contribution!

njouanin:
When do you plan the next release ?

tyarkoni:
Soon! :)

Issue #34, opened 2015-09-01
Logger does not display state information


asobrien:
When using the logging functionality with transitions, the state information of the state machine is not displayed `on_enter` and `on_exit`. For example, on entering a state stdout displays:

```
2015-09-01 13:13:41,528 - transitions.core: Exited state %s
```

This issue is probably due to an incorrect string formatting in calls to the logger, see below:

```python
logger.info("Entered state %s", self.name)

# should be changed to:
logger.info("Entered state %s" % self.name)
```

PR #35, opened 2015-09-01
Fixes string formatting syntax inside logger calls

asobrien:
Closes issue #34.


tyarkoni:
Thanks for fixing this!

PR #44, opened 2015-11-28
Log a warning for an invalid trigger if ignore_invalid_triggers is True

Ch00k:
This adds more verbosity when an invalid transition is executed when `ignore_invalid_triggers` is `True`.


tyarkoni:
Thanks!

PR #47, opened 2015-12-09
Create diagram from machine states and triggers

svdgraaf:
I've split up the graphs, so we can perhaps add different diagram types later on, not sure if we want that right now, but it's easier to scale out later on. The `Machine.get_graph()` method will return a pygraphviz object.

You can render a png like this:
```
In [7]: machine.get_graph().draw('foobar.png', prog='dot')
```
Round Diagram

Or you can render the dot markup directly to string:
```
In [8]: print machine.get_graph()
digraph {
	graph [rankdir=LR,
		ratio=0.3,
		title=AGraph
	];
	node [height=1.2,
		label="\N",
		shape=circle
	];
    ...
}
```

I defaulted to the AGraph diagram type, you can override or provide your own diagram classes by subclassing Diagram or in this case AGraph, eg for creating square images:

```
from transitions.diagrams import AGraph

class BlockAGraph(AGraph):
    state_attributes = {
        'shape': 'square',
        'height': '1.2',
    }

machine.get_graph(diagram_class=BlockAGraph).draw('foobar.png', prog='dot')
```
Look mommy, squares!

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,1 +1,2 @@
 dill
+pygraphviz
--- a/transitions/diagrams.py
+++ b/transitions/diagrams.py
@@ -47,7 +47,7 @@ class AGraph(Diagram):
             shape = self.state_attributes['shape']
 
             # We want the first state to be a double circle (UML style)
-            if state == self.machine.states.items()[0]:
+            if state == list(self.machine.states.items())[0]:
                 shape = 'doublecircle'
             else:
                 shape = self.state_attributes['shape']
--- a/transitions/diagrams.py
+++ b/transitions/diagrams.py
@@ -5,13 +5,14 @@ class Diagram(object):
     def __init__(self, machine):
         self.machine = machine
 
-class AGraph(Diagram)
-    self.state_attributes = {
+
+class AGraph(Diagram):
+    state_attributes = {
         'shape': 'circle',
         'height': '1.2',
     }
 
-    self.machine_attributes = {
+    machine_attributes = {
         'directed': True,
         'strict': False,
         'rankdir': 'LR',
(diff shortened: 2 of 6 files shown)


svdgraaf:
Perhaps we want to add pygraphviz to the default requirements as well? Not sure.

requirements_test.txt
```diff
@@ -1 +1,2 @@
-dill
\ No newline at end of file
+dill
+pygraphviz
```
tyarkoni: Let's add this to tests_require in setup.py as well.

transitions/diagrams.py
```diff
@@ -0,0 +1,63 @@
+import pygraphviz as pgv
+
+
+class Diagram(object):
+    def __init__(self, machine):
+        self.machine = machine
+
+
+class AGraph(Diagram):
```
tyarkoni: Maybe add get_graph to Diagram as an @abstractmethod to make sure it's implemented by any future graphing interfaces.

tyarkoni:
Do you mind adding a test? Doesn't have to be anything fancy; just creating a very simple machine and testing to make sure that the graphing code writes out a file with non-zero size would suffice. I can do it too if you prefer. Otherwise I think this is ready to merge.

svdgraaf:
I'll add a test, good call :)

svdgraaf:
Ok, just need to tweak the python 3 compatibility a bit :D on it!

svdgraaf:
boom :)

tyarkoni:
Awesome, thanks---this is great! I'll update the README.

svdgraaf:
First ever public PR :heart:

tyarkoni:
Congrats! Hope there are many more to come. ;)

PR #52, opened 2015-12-25
Master

aleneum:
added hierarchical (flattened) and reentrant locked machines


aleneum:
Hi,

let me know what you think about this (flattening) approach to HSM and thread-safe function access. I had to do some minor adaptions to the main files to ease subclassing but tried to keep the changes as minimal as possible. The only functional change I have done is to remove the `New` Node from diagram creation and replaced it with the first state in the machines state list. I did this because `New` is not connected to the state diagram if 'New' is not part of the machine. If you find the approach to be sufficient but have some remarks concerning code style/best practise/philosophy, let me know and I will do by best to make the code 'blending' into transitions as best as possible. If this approach does not fulfil some requirements, I am open to constructive criticism/feedback as well :).

Best regards,
Alex

tyarkoni:
This looks great, thanks! I'll review this as soon as I can--probably on Monday.

tyarkoni:
Okay, I've looked through the code... This is really great--thanks so much for the contribution! I only have a few minor stylistic comments (and I'll probably make some minor edits to the README after this is merged):

* I'm not sure about using the underscore character to concatenate nested state names. I think the forward slash is probably just as good for purposes of nesting, and would be less likely to cause confusion, since using the underscore in variable names is quite Python and is likely to be used a lot in states that *don't* involve nesting. I recognize that it probably won't cause any problems functionally if users mix non-nested states with underscores and nested states, but it seems a bit confusing. So unless you have objections, I'd prefer to see 'foo/bar/baz' rather than 'foo_bar_baz'.
* In the counter example, the instance name is 'counter' in some cases and 'calculator' in others--this should be made consistent.
* Siince LockedMachine and HierarchicalMachine seem to have identical code, could we maybe use use a mixin pattern or something to manage this without the redundancy?
* For consistency, we should probably call the LockedHSM class 'LockedHierarchicalMachine'. It's kind of annoying to type, but I think it's good to keep the nomenclature consistent.
* To minimize redundancy, I think the 'Stuff' and 'Inherited Stuff' classes in the testing modules should probably be moved into a separate module (probably 'utils') under testing, and then they can be imported in multiple places as needed.

Otherwise everything looks great! If you don't have time to make some of the above changes, just let me know, and I can always merge first and then patch it myself.

aleneum:
Hi,

thanks for the feedback. 

> In the counter example, the instance name is 'counter' in some cases and 'calculator' in others--this should be made consistent.

fixed

> Since LockedMachine and LockedHSM seem to have identical code, could we maybe use use a mixin pattern or something to manage this without the redundancy?

fixed; I had tried multiple inheritance for `LockedHierarchicalMachine` but could not get it to work. I assumed there were issues with super and `__getattribute__` in multi-inheritance scenarios... well, I tried again, now it works \o/... seems these issues were related to previous locking approaches.

> For consistency, we should probably call the LockedHSM class 'LockedHierarchicalMachine'. It's kind of annoying to type, but I think it's good to keep the nomenclature consistent.

agree and fixed; Co-worker of mine always says 'code is usually read way more often than it is written' and with code completion it does not really matter.

> To minimize redundancy, I think the 'Stuff' and 'Inherited Stuff' classes in the testing modules should probably be moved into a separate module (probably 'utils') under testing, and then they can be imported in multiple places as needed.

fixed; called the module test_utils

> I'm not sure about using the underscore character to concatenate nested state names. 

That one is quite tricky. I started with dots actually (like domains) but went for underscore later on because dots mess with auto_transitions. E.g. model.to_A.1 does not work and model.to_A/1 wont work either. The characters allowed in [Python 2.x keywords](https://docs.python.org/2/reference/lexical_analysis.html#grammar-token-identifier) are quite limited. Double underscores (to_A__1) would be an alternative or replacing '/' or '.' in function keywords only. State would be called 'A.1' but auto_transition would be to_A_1. The first approach stretches functions calls and second approach is rather inconsistent. Both are quite nasty. What do you think?

tyarkoni:
Oh, good point re: auto_transitions. I agree that the alternatives are messier, so let's stick with underscore for now and revisit it in future if it becomes an issue. Thanks for making all the other changes!

wtgee:
On Sat, Jan 2, 2016 at 10:51 AM, aleneum <notifications@github.com> wrote:

> Hi,
>
> thanks for the feedback.
>
> In the counter example, the instance name is 'counter' in some cases and
> 'calculator' in others--this should be made consistent.
>
> fixed
>
> Since LockedMachine and LockedHSM seem to have identical code, could we
> maybe use use a mixin pattern or something to manage this without the
> redundancy?
>
> fixed; I had tried multiple inheritance for LockedHierarchicalMachine but
> could not get it to work. I assumed there were issues with super and
> __getattribute__ in multi-inheritance scenarios... well, I tried again,
> now it works \o/... seems these issues were related to previous locking
> approaches.
>

FYI, I was running into this multiple inheritance issue with v0.2.9 as
well. __getattribute__ was doing a locking check to make sure self !=
self.model but that is gone in favor of the inspect module and I haven't
had problems since then.


> For consistency, we should probably call the LockedHSM class
> 'LockedHierarchicalMachine'. It's kind of annoying to type, but I think
> it's good to keep the nomenclature consistent.
>
> agree and fixed; Co-worker of mine always says 'code is usually read way
> more often than it is written' and with code completion it does not really
> matter.
>
> To minimize redundancy, I think the 'Stuff' and 'Inherited Stuff' classes
> in the testing modules should probably be moved into a separate module
> (probably 'utils') under testing, and then they can be imported in multiple
> places as needed.
>
> fixed; called the module test_utils
>
> I'm not sure about using the underscore character to concatenate nested
> state names.
>
> That one is quite tricky. I started with dots actually (like domains) but
> went for underscore later on because dots mess with auto_transitions. E.g.
> model.to_A.1 does not work and model.to_A/1 wont work either. The
> characters allowed in Python 2.x keywords
> <https://docs.python.org/2/reference/lexical_analysis.html#grammar-token-identifier>
> are quite limited. Double underscores (to_A__1) would be an alternative or
> replacing '/' or '.' in function keywords only. State would be called 'A.1'
> but auto_transition would be to_A_1. The first approach stretches functions
> calls and second approach is rather inconsistent. Both are quite nasty.
> What do you think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/tyarkoni/transitions/pull/52#issuecomment-168427391>.
>



-- 
~Wilfred Tyler Gee

PR #53, opened 2015-12-30
Minor formatting tweaks and edits to README.md

Zearin:


tyarkoni:
Thanks! Most of these changes look great. I generally prefer 'we' to 'you', as it feels more inclusive. But no biggie, so I'll merge as-is.

PR #55, opened 2016-01-03
Diagrams Merged Incorrectly

wtgee:
It looks like the code for the diagrams was merged incorrectly, with `_add_events` being a copy/paste of `_add_states`. I moved the original code for the events into the method and my graphs are generating again but original author might want to check out (@svdgraaf).

### DIFF
--- a/transitions/diagrams.py
+++ b/transitions/diagrams.py
@@ -72,9 +72,7 @@ class AGraph(Diagram):
 
         # For each state, draw a circle
         self._add_nodes(self.machine.states, fsm_graph)
-        # fsm_graph.add_node('null', shape='plaintext', label='')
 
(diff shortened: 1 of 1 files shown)


wtgee:
I'm getting some spurious states in my graph and can't figure out where they come from, so something else needs to be fixed as well. The spurious states each have a single letter as their name but I can't seem to figure out where they are coming from.

wtgee:
Example of extra states:

State Diagram

tyarkoni:
Ah, I see the issue. In _add_nodes(), we have:

```python
state = state[0]
```

...which expects a list of key/item tuples, but is getting the state dict. I'll fix this and also expand the graphing tests. Thanks!

PR #58, opened 2016-01-06
Locked(HSM) Fixes & Diagram Doublecircle

aleneum:
Hi,

I fixed LockedHSM to use LockedNestedTranstions and fixed pickling issues for python 2.7.  I added a pickling test to test_threading. However, pickle does not support thread locks which means it fails now for python > 3 but dill does…

I also ‘fixed’ the double circle thing but could revert changes in `AGraph` and `AAGraph` since @wtgee is also working on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,7 +681,7 @@ machine.to_C_3_a() # exit A, enter C, enter C_3, enter C_3_a
 
 ### Extension: Reuse of previously created HSMs
 
-Besides semantical order, nested states are very handy if you want to specify state machines for specific tasks and plan to reuse them. Transitions offers two ways to achieve this: You can either *pass a state machine* as an argument or retrieve the machine's *blueprint*, store this as plain text somewhere and pass this to a new machine.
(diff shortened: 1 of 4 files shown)


aleneum:
This is where I got this information from:

http://stackoverflow.com/a/23046244/1617563
https://github.com/Koed00/django-q/issues/38#issuecomment-126878806

and this might be a solution:

https://groups.google.com/d/msg/celery-users/E3CurhqxNDI/1dW4Jh0dQMsJ

If pickle support is highly appreciated I'd take a look at the reduce approach.

wtgee:
I'd go ahead and merge the doublecircle from here directly to get it
working.

Cheers,

tyarkoni:
Hmm... I think pickling is important, but I'd rather not hold this PR up until we fix it. But I also don't want to leave the tests failing if we're going to treat this as a known issue that only applies to an edge case (pickling an HSM on >= 3.4). I would suggest (a) removing the pickling test in test_threading (or only running it on Python < 3.4), and (b) opening a new issue for this, which I'll try to get to this weekend if nobody fixes it sooner. Sound good?

tyarkoni:
Thanks!

Issue #59, opened 2016-01-06
Pickling without Dill for Python 3.4 and higher fails for LockedMachine


aleneum:
```python
======================================================================
ERROR: test_pickle (tests.test_threading.TestTransitions)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../transitions/tests/test_threading.py", line 67, in test_pickle
    dump = pickle.dumps(m)
AttributeError: Can't pickle local object 'LockedMachine.__getattribute__.<locals>.locked_method'
```

PR #60, opened 2016-01-07
Fix for #59

aleneum:
To be pickled
* objects should not be local
-> moved locked_method to callable object LockedMethod
* subclasses should not mess around with __getattribute__
-> Even though the method was found with __getattribute__, __getattr__
was called in pickle.load(). I assume this has something to do with the
order in which objects are deserialized. Overriding __getattr__ of the
subclass with __getattribute__ of the base class solves the issue.

I also expanded the pickle test in test_threading to check if locking still works after pickling.


tyarkoni:
Awesome, thanks!

PR #61, opened 2016-01-08
Improved Testing & __getattr__ LockedMachine fix

aleneum:
To reduce redundancy I let test_nesting and test_threading inherit from `TestTransitions` in test_core. This way I stumbled upon a bug in `LockedMachine.__getattr__`. I assumed `__getattribute__` will call `__getattr__` implicitely if it fails to find the item. But that does not seem to be the case. LockedMachine causes an `AttributeError` for         Machine.on_enter_State
Calling `super.__getattr__` as a fallback solves this issue.

### DIFF
--- a/tests/test_graphing.py
+++ b/tests/test_graphing.py
@@ -17,11 +17,14 @@ class TestDiagrams(TestCase):
         transitions = [
             {'trigger': 'walk', 'source': 'A', 'dest': 'B'},
             {'trigger': 'run', 'source': 'B', 'dest': 'C'},
-            {'trigger': 'sprint', 'source': 'C', 'dest': 'D'}
+            {'trigger': 'sprint', 'source': 'C', 'dest': 'D', 'conditions': 'is_fast'},
+            {'trigger': 'sprint', 'source': 'C', 'dest': 'B'}
         ]
 
-        m = Machine(states=states, transitions=transitions, initial='A')
+        m = Machine(states=states, transitions=transitions, initial='A', auto_transitions=False)
         graph = m.get_graph()
+        self.assertIsNotNone(graph)
+        self.assertTrue("digraph" in str(graph))
 
         # Test that graph properties match the Machine
         self.assertEqual(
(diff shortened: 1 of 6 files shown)


tyarkoni:
Sorry for the delay getting to this, @aleneum. Would you mind resolving the merge conflicts on your side? I think at this point, extensions.py is really more your domain. :)

tyarkoni:
Also, @aleneum, I don't see an email address for you in your user profile. Could you please drop me an email (tyarkoni@gmail.com)? Thanks!

aleneum:
Sure, I will do that. Also conflicts are merged and my profile is a bit more informative now.

aleneum:
I found a bug in `diagram.py:_add_edges` where only the first transition triggered for each state is added as edge. Let's assume states `A, B, C, D` and the following transitions:

```python
    transitions = [
        {'trigger': 'walk', 'source': 'A', 'dest': 'B'},
        {'trigger': 'run', 'source': 'B', 'dest': 'C'},
        {'trigger': 'sprint', 'source': 'C', 'dest': 'D', 'conditions': 'is_fast'},
        {'trigger': 'sprint', 'source': 'C', 'dest': 'B'}
]

This should result in four edges but C -> B cannot be found. 

```python
    for transition in event.transitions.items(): # ('C': [<transition sprintCD>, <transition sprintCB>])
        src = transition[0] # C
        dst = transition[1][0].dest # D
        container.add_edge(src, dst, label=label)
```

`sprintCB` is not considered. I changed this to:

```python
for transitions in event.transitions.items():
    src = transitions[0]
    for t in transitions[1]:
        dst = t.dest
        container.add_edge(src, dst, label=label)
```

`test_graphing` was extended with an edge count test. I also moved `self.assertIsNotNull(graph)` a bit up in `test_graphing.py` since graph.get_edges() was called before this test.

tyarkoni:
Looks good to me, and thanks catching the bug.

PR #62, opened 2016-01-12
Change some log messages' level from info to debug

Ch00k:
Reduces the amount of logging a bit. Especially useful when transitions is used together with another module that has its own logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,14 +605,14 @@ Here you get to consolidate all state machine functionality into your existing m
 
 ### Logging
 
-Transitions includes very rudimentary logging capabilities. A number of events--namely, state changes, transition triggers, and conditional checks--are logged as INFO-level events using the standard Python `logging` module. This means you can easily configure logging to standard output in a script:
+Transitions includes very rudimentary logging capabilities. A number of events -- namely, state changes, transition triggers, and conditional checks -- are logged using the standard Python `logging` module (with different log levels). This means you can easily configure logging to standard output in a script:
 
 ```python
 
 # Set up logging
 import logging
 from transitions import logger
-logger.setLevel(logging.INFO)
+logger.setLevel(logging.INFO)  # logging.DEBUG for verbosity
 
 # Business as usual
 machine = Machine(states=states, transitions=transitions, initial='solid')
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -48,14 +48,14 @@ class State(object):
         for oe in self.on_enter:
             event_data.machine.callback(
                 getattr(event_data.model, oe), event_data)
-        logger.info("Entered state %s" % self.name)
+        logger.debug("Entered state %s" % self.name)
 
     def exit(self, event_data):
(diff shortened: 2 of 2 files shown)


Ch00k:
A change in README.md decreased coverage by 0.3%?

tyarkoni:
This change seems sensible to me, but let's give it a day or two for other people to weigh in with any possible objections, since it does break behavior that some users may have come to expect.

Not sure what the change in coverage is about (might be because it's scanning README for docstrings or something?), but no need to worry about it.

aleneum:
Hi,

I googled around a bit to find some accepted loglevel guidelines about logging but could not find THE guideline. The essence of what I have read so far is  "`INFO` is for operators and `DEBUG` is for developers". Concerning this, I'd suggest to keep the transition and execution messages at the `INFO` level since it tells you as a user of the package what it is doing right now. But that's just my point of view.

If that does not work well with your code and your global log level, I'd suggest to set the level for the 'transitions' domain specifically like 
```python
logging.getLogger("transitions").setLevel(logging.WARNING)
```
That's how I deal with verbose libraries and packages.

About changing the failed condition check level from `INFO` to `WARNING`:
```python
logger.error("Transition condition failed: %s() does not " + 
     "return %s. Transition halted.", c.func, c.target)
```
A "failed" condition check is part of the standard workflow and should not cause an error message.
Think about a polled transition `{name: 'upshift', 'source': 'Gear1', dst:'Gear2', 'conditions':'fast_enough'}` which checks periodically if a car should change gear. Or a "if X go to StateA else go to StateB" case where the condition check of A will and should fail from time to time.

tyarkoni:
On thinking about it some more, I think I agree with @aleneum here. So let's leave logging as-is for now. At some point it would be nice to beef up the logging functionality (e.g., by standardizing the format of the output in some machine-readable format, and adding graphing of transition history). At that point we could conceivably add some basic filtering tools, so that one can imagine calling something like logger.get_transitions() vs. logger.get_events(), where the former would return only successful transitions, and the latter returns all events of potential interest.

In the shorter term, I'm certainly willing to consider a revision to the logged codes that can be more easily filtered post-hoc by the user. E.g., instead of a human-readable message like "Transitioned from state S1 to state S2", it might make more sense to log something like "TRANSITION\tS1\tS2\tSUCCESS\n", "CALLBACK\tS1\tafter_S1\n", and so on. Then it would be trivial to grab only the TRANSITION codes with SUCCESS, etc.

Ch00k:
@aleneum @tyarkoni I agree with your remarks. What I was trying to achieve with this change is indeed quite specific to my use case.
Commit cf894732f1dabd5bd11c2eab1441d9341a75a38c though is still important as it fixes an issue introduced by my other MR earlier (sorry for that). Perhaps the easiest is to just commit it directly, without the MR.

tyarkoni:
Sounds good, thanks @Ch00k. I made that change directly.

PR #68, opened 2016-01-19
Dynamic diagram nodes & edges

wtgee:
This adds the ability to store the graph object along with the machine so that it can be updated dynamically. Two public methods are provided to update the nodes and the edges. Three state styles added.

Example of how to add auto-update to model class:
```python
def __init__(self):
        Machine.__init__(
            self,
            states=_states,
            transitions=_transitions,
            initial=_initial
            send_event=True,
            before_state_change='before_state',
            with_graph=True,
        )

        self._previous_state = None
        self._next_state = None
        self._update_graph(with_previous=False)

def before_state(self, event_data):
    """ Called before each state.   """
    self._next_state = event_data.event.transitions.get(self.state)[0].dest
    self._previous_state = event_data.state.name
    self._update_graph()

def _update_graph(self, with_previous=True):
    """ Show the active state on the graph """
    if self._next_state is None:
        self._next_state = self.state

    # Mark the active node
    self.set_node_state(self._next_state, state='active', reset=True)

    # Mark the previous node and path used
    if with_previous:
        if self._previous_state is not None:
                self.set_node_state(self._previous_state, state='previous')
                self.set_edge_state(self._previous_state, self._next_state, state='previous')
```

* `Machine` option to store the graph
* `set_node_state` and `set_edge_state` work with stored graph to update
    appearance.
* Added 'default', 'active', and 'previous' state attributes

### DIFF
--- a/tests/test_graphing.py
+++ b/tests/test_graphing.py
@@ -70,7 +70,50 @@ class TestDiagrams(TestCase):
         for t in triggers:
             self.assertIsNotNone(getattr(m, t))
 
-        self.assertEqual(len(graph.edges()), 8) # see above
+        self.assertEqual(len(graph.edges()), 8)  # see above
+
+        # write diagram to temp file
+        target = tempfile.NamedTemporaryFile()
+        graph.draw(target.name, prog='dot')
+        self.assertTrue(os.path.getsize(target.name) > 0)
+
+        # cleanup temp file
+        target.close()
+
+    def test_store_nested_agraph_diagram(self):
+        ''' Same as above, but with nested states. '''
+        states = ['A', 'B', {'name': 'C', 'children': ['1', '2', '3']}, 'D']
+        transitions = [
+            {'trigger': 'walk', 'source': 'A', 'dest': 'B'},   # 1 edge
+            {'trigger': 'run', 'source': 'B', 'dest': 'C'},    # + 1 edge
+            {'trigger': 'sprint', 'source': 'C', 'dest': 'D',  # + 3 edges
+             'conditions': 'is_fast'},
+            {'trigger': 'sprint', 'source': 'C', 'dest': 'B'}  # + 3 edges = 8 edges
+        ]
+
+        m = HierarchicalMachine(
+            states=states, transitions=transitions, initial='A', auto_transitions=False, with_graph=True)
+        graph = m.get_graph()
+        self.assertIsNotNone(graph)
+        self.assertTrue("digraph" in str(graph))
+
+        # Test that graph properties match the Machine
+        # print((set(m.states.keys()), )
+        node_names = set([n.name for n in graph.nodes()])
+        self.assertEqual(set(m.states.keys()) - set('C'), node_names)
+
+        triggers = set([n.attr['label'] for n in graph.edges()])
+        for t in triggers:
+            self.assertIsNotNone(getattr(m, t))
+
+        self.assertEqual(len(graph.edges()), 8)  # see above
+
+        # Force a new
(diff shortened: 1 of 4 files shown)


wtgee:
Example of active and previous states marked:

PSM

aleneum:
Neat! looking forward to this feature! how 'fast' can you update graphs with this? close to realtime?

Small bug
```python
# core.Machine
def get_graph(self, title=None, force_new=False, diagram_class=AGraph):
# extensions.HierarchicalMachine
def get_graph(self, title=None, force_new=False, diagram_class=AAGraph):
    if self._graph is None or force_new:
        # misses a parameters; diagram_class is mapped to force_new
        self._graph = super(HierarchicalMachine, self).get_graph(title, diagram_class)
       # should be  self._graph = super(HierarchicalMachine, self).get_graph(title, force_new, diagram_class)
```

After that, the tests pass as well.
Best regards,
Alex

aleneum:
Some thoughts about the usage:

* Rather than relying on `send_event=True`, I'd suggest to call _update_graph implicitly if `with_graph=True`. You could subclass/mixin Transition instead for instance. `send_event=true` may require complete rewrites of existing models and also may lead to less comprehensible code since you pass event_data which is not necessarily required.
`Model.speed_reached(velocity)` is easier to read and debug than `Model.speed_reached(event_data)` with checks like  `if hasatttr(event_data, velocity)`

* This line is problematic:
` self._next_state = event_data.event.transitions.get(self.state)[0].dest`
There is no guarantee that the first transition will be executed. Another reason why you might want to connect `_update_graph` and the transition which will be executed.

Suggestion:
* make get_graph a mixin
```python

class MachineGraphSupport(object):
    def get_graph(self, title=None, force_new=False, diagram_class=AGraph)
    def set_edge_state(self, edge_from, edge_to, state='default')
    def set_node_state(self, node_name=None, state='default', reset=False)

class TransitionGraphSupport(object):
       # in _change_state it is certain that the transition will happen
       def _change_state(self, event_data):
           event_data.machine.set_edge_state
           event_data.machine.set_node_state
           # this will only work if this class is used as mixin
           super(TransitionGraphSupport, self)._change_state(event_data)
```
This would work without `with_graph`. In this case I'd also rework `LockedMachine` and `HierarchicalMachine` a bit to be pure mixins themselves.
Users can mix their Machines explicitly:

```python
class LockedHierarchicalGraphMachine(LockedMachine, MachineGraphSupport, HierarchicalMachine)
```

Or there are already available in the library or provide a machine class factory which also creates the appropriate transition... 

What do you think?

wtgee:
I agree with a lot of your ideas. Mostly I had this for my model but thought it might be nice to be an actual part of the `Machine` class. If we think it is something nice we can integrate in the manner that you suggest, which I agree is a lot better to do.

All I am really doing is storing the graph object and then updating certain nodes. My `Machine`s have all been fairly small (< 15 states) so updating/drawing takes almost no time at all. The alternative approach would be to just generate a SVG and then update that elsewhere but this proved to be easier.

One annoying thing is that the graph redraws itself based on the colors, so sometimes state nodes move around. I'm not sure if there is a way to fix that or not.

So is this a feature we'd like to add, assuming it gets cleaned up?

aleneum:
I am all for it :+1:. What about if you integrate this feature in a dev-branch in this git and I will refactor the code in extension modules and mixins. After that we review the changes and create a pull request from there if there is no more issue to be clarified.

wtgee:
Agreed. I'm going to close this PR and get a more formal solution going in a separate branch. I'll let you know when it's up. Thanks!

PR #70, opened 2016-01-23
fixed get_state() documentation

avikam:
Hi, 
really small change :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ lump.is_gas()
 >>> False
 lump.is_solid()
 >>> True
-machine.get_state().name
+machine.get_state(lump.state).name
 >>> 'solid'
 ```
 


wtgee:
Perhaps we should move the `get_state` example to the __States__ section and change this example to
`machine.current_state.name`. Or at least also include the `current_state`, which currently isn't documented anywhere.

wtgee:
I made a few small changes in my branch, let me know what you think:

https://github.com/wtgee/transitions/tree/metadata_changes#checking-state

wtgee:
I'm going to merge this just because this corrects an obvious mistake. I improved the example a bit with the `.current_state` attribute (as shown in link above) but will just merge those changes in with some other project metadata changes. Thanks again for catching this @avikam

PR #72, opened 2016-01-24
Very minor text change

SteveAmor:
Helps when read for the first time if you know the verbose declaration of transition, the first time it is used.


aleneum:
:+1:. I will pull it if there no objection until tomorrow.

wtgee:
:+1: Makes sense to me.

PR #73, opened 2016-01-24
this fixes #67; event.trigger is locked rather than single transitions.

### LINKED ISSUE #67: Lock fails for conditional transition
I just stumbled upon a bug which occurs if a list of possible transitions is evaluated in `core.Event`. This problem occurs because locking `core.Transition.execute` is not sufficient for the test scenario below:

```python
def test_conditional_access(self):
    self.stuff.heavy_checking = heavy_checking # checking takes 1s and returns False
    self.stuff.machine.add_transition('advance', 'A', 'B', conditions='heavy_checking')
    self.stuff.machine.add_transition('advance', 'A', 'C')
    t = Thread(target=self.stuff.advance)
    t.start()
    time.sleep(0.1)
    logger.info('Check if state transition done...')
    # Thread will release lock before Transition is finished
    self.assertTrue(self.stuff.machine.is_state('C')) 
```

Cause:

```python
for t in self.transitions[state_name]:
    event.transition = t
    if t.execute(event): # lock is acquired and released here
```
Execution order will be TransitionAB.execute (Thread), Machine.is_state (Main), TransitionAC.execute (Thread). Update: Since it is a racing condition who gets the lock after TransitionAB it does not fail all the time.

Proposed solution: Lock `core.Event.trigger` rather than `core.Transition.execute`.
~~This requires minor changes to `core` to be able to subclass Event.~~ Update: It does not. It's is three lines of redundant code though.

If you do not object and/or think this is a bad idea I will fix it in a branch `fix-67-threading`.

aleneum:
See #67 for detailed information.

### DIFF
--- a/tests/test_threading.py
+++ b/tests/test_threading.py
@@ -5,6 +5,7 @@ except ImportError:
 
 import time
 from threading import Thread
+import logging
 
 from transitions import LockedHierarchicalMachine
 from transitions import LockedMachine
@@ -18,8 +19,17 @@ except ImportError:
     from mock import MagicMock
 
 
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+
 def heavy_processing():
     time.sleep(1)
+
+
+def heavy_checking():
+    time.sleep(1)
+    return False
 
 
 class TestLockedTransitions(TestCore):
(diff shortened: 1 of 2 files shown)


aleneum:
If there is not objection, I'd like to merge the fix tomorrow (Tuesday).

wtgee:
:+1: Looks good to me, thanks.

tyarkoni:
Me too!

Issue #40, opened 2015-11-11
Before and after state change


omago:
Hi everybody,

I noticed one thing, "condition" callback is triggered before "before" callback. I expected that "before" is called before "condition"? Does my logic make any sense?

tyarkoni:
The issue here is that people can reasonably have different expectations about what should happen if a condition fails. One view is that a "before" callback should always run, even if the transition never completes successfully (because a condition fails). The counterargument is that you might not want before callbacks to fire at all if you know the transition won't happen, because it could leave you in an internally inconsistent state. For a lightweight package like this, my sense is that the latter case is probably the behavior most users expect. So I implemented it with condition-checking first, to prevent any further logic in the event the transition is invalidated.

I'd be happy to change the order if there's general agreement that that's preferable, but I worry that that would introduce undesirable behavior in other respects. An unambiguous way to resolve this would be to split the before callback into two separate callbacks (e.g., 'before' and 'around'), but I'm kind of hesitant to keep adding callbacks as my overriding concern is to keep it simple.

Labeling this as a question so other people can weigh in.

omago:
Thanks for the reply. My view is that before callback is some kind of preparation stage for the transition, as such my view is that it need to be triggered always. After before "condition" is triggered and "condition "is the one that decides about successful transition. After that "after" callback is triggered. Also it would be cool to have two after callback, after_success, after_fail - something like that.

tyarkoni:
That's reasonable. I'll leave this open for further comment, and if there's enough sentiment in favor of always firing "before" events, we can do it that way in a future release.

harshitjain-cs:
Is there any way to pass data along with defining conditions that use that data, in a single transition?

harshi3t:
Is there any way to pass data along with defining conditions using the same data, in a single transition?
something like : 
`machine.add_transition('trigger', 'stateA', 'stateB', before=['set_environment', 'fn1'], conditions=['cond1'])`
Here fn1 and cond1 use the data passed through trigger

tyarkoni:
Yes, conditions and before/after callbacks all receive EventData just like other callbacks. See the "Passing Data" section in the README. If it doesn't work as described, open a new issue.

aleneum:
Hi,

I do not want to judge who is right or wrong here but for me "before" refers to the transition which is about to occur rather than the check if the transition should occur or not. I am pretty sure that there are cases where you prefer to reuse the preparation of the check but the same is true for (certain) transitions. 

[...]

PR #78, opened 2016-02-01
Modify travis config to use container builds

TheMysteriousX:
While poking around with #40 I made a few tweaks to the readme for consistency, and changed the travis config because 3-minute-long builds are an awful thing. I've split these out separately from my other branches as they're probably useful.

The new container infrastructure is much faster - there are a few caveats (can't use sudo is the major one) but none that would affect transitions that I can see.

The docs are below; please check them before merging in case I've missed anything. Travis-CI are happy enough with the new infrastructure that it has been the default for new repositories for a year.

This link: https://docs.travis-ci.com/user/workers/standard-infrastructure indicates the change can be backed out by reverting the .yml and adding `sudo: required`.

https://docs.travis-ci.com/user/migrating-from-legacy/
https://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ python:
   - "3.4"
   - "3.3"
   - "2.7"
-before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -y graphviz
 install:
   - pip install python-coveralls coverage
   - pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


tyarkoni:
Looks good to me, thanks!

PR #80, opened 2016-02-05
"prepare" callback

TheMysteriousX:
There seemed to be something of a consensus forming in #40, so I've continued with the changes I've already made by renaming the 'action' callback to 'before_check'.

I've also added some compatibility code, which is maybe a bit coarse but should work to keep older code working.

Reworking __getattr__ was a little annoying, but it didn't correctly handle underscores.

I'm not 100% certain this is ready to be merged, I'm just soliciting feedback at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,39 @@ lump.evaporate()
 >>> "where'd all the liquid go?"
 ```
 
+There is also a `'prepare'` callback that is executed as soon as the trigger runs, before any `'conditions'` are checked:
+
+```python
+class Matter(object):
+    heat = False
+    attempts = 0
+    def count_attempts(self): self.attempts += 1
+    def is_really_hot(self): return self.heat
+    def heat_up(self): self.heat = random.random() < 0.25
+    def stats(self): print('It took you %i attempts to melt the lump!' %self.attempts)
+
+states=['solid', 'liquid', 'gas', 'plasma']
+
+transitions = [
+    { 'trigger': 'melt', 'source': 'solid', 'dest': 'liquid', 'prepare': ['heat_up', 'count_attempts'], 'conditions': 'is_really_hot', 'after': 'stats'},
+]
+
+lump = Matter()
+machine = Machine(lump, states, transitions=transitions, initial='solid')
+lump.melt()
+lump.melt()
+lump.melt()
+lump.melt()
+>>> "It took you 4 attempts to melt the lump!"
+```
+
+In summary, callbacks on transitions are executed in the following order:
+
+* `'prepare'` (executed as soon as the trigger is called)
+* `'conditions'` / `'unless'` (conditions *may* fail and halt the transition)
+* `'before'` (executed while the model is still in the source state)
+* `'after'` (executed while the model is in the destination state)
+
 ### Passing data
 Sometimes you need to pass the callback functions registered at machine initialization some data that reflects the model's current state. Transitions allows you to do this in two different ways. 
 
--- a/README.md
+++ b/README.md
@@ -494,7 +494,7 @@ lump.evaporate()
 >>> "where'd all the liquid go?"
 ```
 
-There is also a `'prepare'` callback that is executed as soon as the trigger runs, before any `'conditions'` are checked:
+There is also a `'prepare'` callback that is executed as soon as a transition starts, before any `'conditions'` are checked or other callbacks are executed.
 
 ```python
 class Matter(object):
@@ -520,9 +520,11 @@ lump.melt()
 >>> "It took you 4 attempts to melt the lump!"
 ```
 
+Note that if the transition is not valid from the current state, or there are multiple transitions sharing a trigger, `'prepare'` will only be called on the ones that are actually executed.
+
 In summary, callbacks on transitions are executed in the following order:
 
-* `'prepare'` (executed as soon as the trigger is called)
+* `'prepare'` (executed as soon as the transition starts)
 * `'conditions'` / `'unless'` (conditions *may* fail and halt the transition)
 * `'before'` (executed while the model is still in the source state)
 * `'after'` (executed while the model is in the destination state)
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -151,6 +151,35 @@ class TestTransitions(TestCase):
         m.model.move()
         self.assertEquals(m.model.level, 3)
 
+    def test_prepare(self):
+        m = Machine(Stuff(), states=['A', 'B', 'C'], initial='A')
+        m.add_transition('move', 'A', 'B', prepare='increase_level')
+        m.add_transition('move', 'B', 'C', prepare='increase_level')
+        m.add_transition('move', 'C', 'A', prepare='increase_level', conditions='this_fails')
+        m.add_transition('dont_move', 'A', 'C', prepare='increase_level')
+
+        m.prepare_move('increase_level')
+
+        m.model.move()
+        self.assertEquals(m.model.state, 'B')
+        self.assertEquals(m.model.level, 3)
+
+        m.model.move()
+        self.assertEquals(m.model.state, 'C')
+        self.assertEquals(m.model.level, 5)
+
+        # State does not advance, but increase_level still runs
+        m.model.move()
+        self.assertEquals(m.model.state, 'C')
+        self.assertEquals(m.model.level, 7)
+
+        # An invalid transition shouldn't execute the callback
+        with self.assertRaises(MachineError):
+            m.model.dont_move()
+
+        self.assertEquals(m.model.state, 'C')
+        self.assertEquals(m.model.level, 7)
+
     def test_state_model_change_listeners(self):
         s = self.stuff
         s.machine.add_transition('go_e', 'A', 'E')
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -369,3 +369,131 @@ class TestTransitions(TestCase):
         m2 = pickle.loads(dump)
         self.assertEqual(m.state, m2.state)
         m2.run()
+
+    def test___getattr___and_identify_callback(self):
+        m = Machine(Stuff(), states=['A', 'B', 'C'], initial='A')
+        m.add_transition('move', 'A', 'B')
+        m.add_transition('move', 'B', 'C')
+
+        callback = m.__getattr__('before_move')
+        self.assertTrue(callable(callback))
+
+        with self.assertRaises(MachineError):
+            m.__getattr__('before_no_such_transition')
+
+        with self.assertRaises(MachineError):
+            m.__getattr__('before_no_such_transition')
+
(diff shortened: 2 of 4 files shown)


aleneum:
looks good so far.

Some thoughts:

`Machine.add_transition` and `HierarchicalMachine.add_transition` differ concerning keywords or let's say `HierarchicalMachine.add_transition`does not include `before` and `before_check`. Not sure how python handles this. My first guess is that this breaks overriding in cases where `before_check` is present.

I'd prefer a raised error instead of 'automatic fixing' since altering callbacks maybe complicate bug tracking. If the method is named `after` this leads to an error in `name.split(separator)[1]` I guess. Or if it is called `on_exit` it becomes `before_transition_exit`. I know this is heavy abuse of the method but you never know what people come up with :).

And -- I know its a drop in the bucket -- maybe make `separator` and `callbacks` class variables since they are more or less static and `Machine._identify_callback` is probably called frequently. Saves two assignments (weeeee).

TheMysteriousX:
> HierarchicalMachine.add_transition does not include before and before_check
I'll address this; I thought that the nested stuff hadn't been hadn't been included a release so didn't add `before`; the missing `before_check` is an oversight.

> I'd prefer a raised error instead of 'automatic fixing' 
Can do, that makes things much neater too.

> If the method is named after this leads to an error in name.split(separator)[1]
I'll add a regression test and fix accordingly.

> make separator and callbacks class variables
As they're constants, the interpreter should optimise them out by in-lining the values where needed. I'll double check what bytecode is actually produced.

TheMysteriousX:
* The signature inconsistency has been fixed.
* After thinking about it for a while I realised that the naming of callbacks is completely deterministic; even if someone names a transition something like "on_exit_A", it will be correctly handled with the changes in this version, so no exceptions or even log messages (beyond the deprecation warning) needed.
* In the previous version the constants were indeed optimised out; they're now operated on directly though (cls.seperator.join etc.), so I've moved them to class scope to avoid them being assigned multiple times.

aleneum:
> In the previous version the constants were indeed optimised out;

That's good to know. I haven't spent much thoughts about python optimisation. But it makes sense.

> so no exceptions or even log messages (beyond the deprecation warning) needed.

I still have comprehension issues with the 'autofixing':
```python
if len(name) == len(callback_type) and callback_type in ['before_transition', 'before_check']:
    logger.info('"before" callback is deprecated; use "before_transition_*" (callback was: "%s")', name)
    name = separator.join(['before_transition'] + name.split(separator)[1:])
```
This is triggered if `name` equals `'before_transition'` or `'before_check'`. Since `callback_type` has to be a substring of `name` (otherwise the `ValueException` would have caused a return), the length comparison is basically a check for equal value (or isn't it?).
This results in outputs like `before_transition_transition` or `before_transition_check`. Is that the desired behaviour?

TheMysteriousX:
That section fixes an edge case where a user has created a transition called `transition` or `check`.

In this case, you have the `callback_type` being extracted as either `before_transition` or `before_check` with no remainder - `name` and `callback_type` will be equal.

This could be rewritten like this:

```
if name in ['before_transition', 'before_check']
```

But there is a performance hit as in all code paths you're doing a full string comparison.

> This results in outputs like before_transition_transition or before_transition_check. Is that the desired behaviour?

Yes - the case we're cleaning up is:

```
Machine.add_transition('check')
Machine.before_check('do_this_callback)  # __getattr__ maps this to before_transition_check
```

Is that clearer?

aleneum:
> But there is a performance hit as in all code paths you're doing a full string comparison.

oh, I get it. I'd suggest just a adding a comment in the code and leave it like it is. Then it should be clear why this is the better solution.
> Yes - the case we're cleaning up is

understood. The case is transparent. It just causes issues of somebody really messes with the system and has a `check` and `transition_check` transition for instance. That's very unlikely and I'd take the risk. @wtgee: what do you think?

aleneum:
I pulled your changes into `feature-#40` branch to resolve the minor conflicts which were added in #82. 
I also wrote a test for the described case:
```python
def test_autofixing_before(self):
    self.stuff.before_check_callback = MagicMock()
    self.stuff.machine.add_transition('check', 'A', 'B')
    self.stuff.machine.before_check('before_check_callback')
    self.stuff.check()
    self.assertTrue(self.stuff.machine.is_state('B'))
    self.assertTrue(self.stuff.before_check_callback.called)
```
It throws an `AttributeError`. Fix in `Machine._identify_callback`:
```python
if len(name) == len(callback_type) and callback_type in ['before_transition', 'before_check']:
    logger.info('"before" callback is deprecated; use "before_transition_*" (callback was: "%s")', name)
    name = cls.separator.join(['before_transition'] + name.split(cls.separator)[1:])
    callback_type = 'before_transition' # <--- if case misses this line
```
Without it target becomes `[before_trans]ition_check`. I guess that is not intended.

But I have seen this int `test_core.py`
```python
def test_callback_identification(self):
    m = Machine(Stuff(), states=['A', 'B', 'C', 'D', 'E', 'F'], initial='A')
    ...
    m.add_transition('check', 'C', 'E', before='increase_level')
    ...
    with self.assertRaises(AttributeError):
        m.before_check('increase_level')
```

This implies that `before_check` is not valid. This contradicts the previously mentioned if statement's purpose, doesn't it?

TheMysteriousX:
You're correct about both of those - looks like I didn't push my branch with my last set of changes (I've done it now).

I had removed that no longer valid case and replaced it with these two: https://github.com/tyarkoni/transitions/pull/80/files#diff-cc4345db19ff44863e9122c74e9f383fR401

Your fix there is what I have.

aleneum:
okay, merged the changes into the branch. (Probably) last question:
your commit comments read like using class variables in `cls.callbacks[[name.find(x) for x in cls.callbacks].index(0)]` cannot be optimized like the previous `callbacks = ['before_transition', 'after', 'before_check', 'on_enter', 'on_exit', 'before']`. Should we change it back then?

TheMysteriousX:
No, in this form using cls is more efficient (by about half a second over 20 million iterations). I didn't dig into the reason too deeply but I suspect it's because:

```
cls.separator.join(['before_transition'] + name.split(cls.separator)[1:])
```

Makes a call against `separator` directly, so it has to be assigned rather than embedded as a constant.

aleneum:
alright, I have opened a pull request to finally integrate your work into the master under #84.

wtgee:
Hi everyone, sorry for the delay but I finally had to a chance to dig around here. It looks like a nice clean implementation (and I'm glad for the `__getattr__` rework). 

The naming to me is a little weird. I'd prefer something like `prepare` instead of `before_check` for the pre-condition and then just a simple `before` and `after` as it was. I think that makes more sense in my head but I'm not tied to it (and more people will probably have implementation methods called `prepare` than `before_check`). Maybe even `before_condition` would make more sense but that doesn't help with the verbosity.

Also, just for my clarification, if a `condition` fails is the user expected to do any cleanup for that in the next `before_check` call? To me that seems like the `before_check` would then have to be fairly generic rather than `state` (or `transition`) specific. If that is the case, it seems like a general `system_check` that happens before any conditions would accomplish the same thing.

aleneum:
> Maybe even before_condition would make more sense

I guess `before_condition` may be more consistent. Let's see what the others say. This change can be done in no time.

> To me that seems like the before_check would then have to be fairly generic rather than state (or transition) specific.
It is in some sense transition specific since it is guaranteed to be executed if the transition is valid.

Right now execution order is before_check --> check -false-> check -false-> ...
There is no after_check/condition implemented (yet). This means `before_check` must not require cleanup. Otherwise the chain of responsibility is broken and every condition check has to execute pretty similar code. 

-- further thoughts

I guess most of the work was making the underscore processable in `__getattr__ `. So  `after_condition` is not that hard to add but then we run into the execution order problem:
* chain execution: before_check/condition -> check -> check -> after_condition
* single execution: before_check/condition -> check -> after_condition -> before_check -> check -> after_condition

We could tackle this with a strategy keyword (e.g.: 'single_execution=False') but imho this sounds like way too much to configure for a user considering the initial goal was just to execute some code every time an event was triggered.

tyarkoni:
+1 for prepare. I know I was the one who pushed for `before_check` and `before_transition` in the first place (sorry!0, but on reflection, `prepare` and `before` probably work better. `prepare` nicely captures the intended semantics, and also gets us out of having to deal with any deprecation issues, since we wouldn't be breaking the API at all.

I'm ambivalent about adding a general `system_check` callback. It would be nice to have, but is another one of those places where once you start down that road, you can add a lot of other features pretty quickly, and it can get hard to keep track of all the functionality. If we do want to pursue this, another avenue I'd prefer to explore--which would add more code, but probably be cleaner in the long run--is to abstract the callback functionality, so that we can easily insert new callbacks by doing something like `check_callbacks('before`, event_data)` at various places in the code, instead of looping over hardcoded named callbacks. It gets kind of ugly having `self.before_check`, `self.before_transition`, etc. transitions... would be cleaner to just store a `callbacks` dict inside the model, and I think it would work just as well and improve code maintainability. I don't think we should tackle this right away, but it's something to keep in mind. Given that kind of implementation, I'd feel much better about adding additional callbacks, since it would add almost no new complexity each time.

adam-bishop:
I'm happy to change it to `prepare` - I've got some time this evening so I'll rebase my changes with the different names, then apply @aleneum 's changes on top where they're still required.

tyarkoni:
Oh, and forgot to mention--the patch looks great to me otherwise; thanks @TheMysteriousX!

TheMysteriousX:
I've renamed `before_check` to `prepare`, reverted the `before` rename and stripped the compatibility code.

Everything has been rebased onto the current master, so the changeset is nice and clean. It could probably do with someone else giving it a thorough check before merging, in case I went too crazy stripping bits out.

From @aleneum's additional changes in #84:

* c4fd30b, 2bcf17c, 1fd7ec6
Merge commit, git fast forwarded over.

* a3cfc2c
No longer needed as there is no longer a collision.

* 7813c5d
Manually merged, I'm not sure if the behaviour is right in the original though. Should `before_state_change` be added to `prepare` (line 214)? I've assumed not.

aleneum:
Looks good to me,

minor doc nitpicking: `prepare` is only called if the transition is evaluated. It is not executed if:
* the transition is invalid (wrong source)
* a previous transition is executed (a transition added *before* the transition in question since at the moment addition order also implies priority)

Not sure if someone will actually assume that the code is *always* executed if the trigger fits.

The blueprint mechanism will (hopefully) be removed pretty soon. So your changes to `nesting.py` are sufficient. If @tyarkoni and @wtgee do not object I will merge your changes into `master` tomorrow if they haven't done it by then :). Thanks for your contribution!

TheMysteriousX:
> minor doc nitpicking

Fair comment, I'll push a quick fix.

Issue #82, opened 2016-02-08
Callables as callbacks


bennyrowland:
At the moment callbacks are set using strings which reference methods on the model object. This has the advantage that a set of transitions is model agnostic and could be used on multiple state machines, however in some cases it would be useful to pass a callable (like a function) as a callback. I have a use case at the moment where the model class has objects as properties, I want to call methods on them and currently I have to wrap those methods in methods defined on the model class itself. This means I have twice as many functions just to pass the callbacks around. I think it would be relatively straightforward to maintain the current usage and just do a type comparison when the callbacks need to be called, if they are strings look up the attribute on the model, if they are callables call them directly.

tyarkoni:
+1 for this. It would definitely be nice to allow users to pass callables, and should be pretty easy to patch in. I'm happy to take a pass at this, though it probably won't happen until next weekend. If you want to work up a PR before then, that would be great.

aleneum:
How about that -> #83

PR #83, opened 2016-02-09
allow callables for callbacks

aleneum:
see #82 for details


tyarkoni:
Very elegant! Thanks. :)

PR #84, opened 2016-02-10
before_check and before_transition Integration

aleneum:
This branch integrates the work of @TheMysteriousX into transitions. Discussion can be found here: #40 and integration progress here: #80.

@tyarkoni, @wtgee: since that is a major change of how transitions works (especially under the hood) I wait for your okay before I pull the trigger.


aleneum:
rework of #80 makes this obsolete

PR #85, opened 2016-02-12
diagrams: show conditions on edge labels

khigia:
On diagrams, change edge label to show the condition, e.g. 'heat / is_flammable'.

Conditions are separated with '&' and "unless conditions" are preceded by '!', e.g. 'heat / !is_flammable & !is_really_hot'.

### DIFF
--- a/transitions/extensions/diagrams.py
+++ b/transitions/extensions/diagrams.py
@@ -80,7 +80,7 @@ class AGraph(Diagram):
                     container.add_edge(src, dst, label=lbl)
 
     def _transition_label(self, edge_label, tran):
-        if tran.conditions:
+        if self.machine.show_conditions and tran.conditions:
             return edge_label + AGraph.trigger_condition_sep + '&'.join(
                 c.func if c.target else '!' + c.func
                 for c in tran.conditions
@@ -168,13 +168,15 @@ class MachineGraphSupport(Machine):
         self.set_node_style(self.graph.get_node(self.current_state.name), 'active')
 
     def __init__(self, *args, **kwargs):
-        # remove title from keywords
+        # remove graph config from keywords
         title = kwargs.pop('title', 'State Machine')
+        show_conditions = kwargs.pop('show_conditions', False)
         super(MachineGraphSupport, self).__init__(*args, **kwargs)
 
         # Create graph at beginnning
-        self.graph = self.get_graph(title=title)
+        self.show_conditions = show_conditions
         self.title = title
+        self.graph = self.get_graph(title=title)
 
         # Set initial node as active
         self.set_node_style(self.graph.get_node(self.initial), 'active')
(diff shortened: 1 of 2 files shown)


aleneum:
Hi @khigia,

interesting addition. Could you make this an optional feature by adding a keyword (e.g. `show_conditions`) to MachineGraphSupport and `get_graph`. `show_conditions` could be handled like the `title` keyword.

khigia:
Added a parameter show_conditions to make this behaviour optional (and default is disable).

khigia:
Hum ... sorry, will change again, to follow UML-like notation: `transition [condition] / action`

aleneum:
neat, could you attach an image of an example graph with this features enabled?

wtgee:
Hi @khigia, just wanted to follow-up on this issue. If you can, will you post a pic of the graph with this option enabled and then we can get this merged.
Thanks!

khigia:
Sure! Sorry, missed the comment asking for pic :(.

Here is code example and result in attached png file:
```python
from transitions.extensions.diagrams import MachineGraphSupport

fsm = MachineGraphSupport(
    states=['A', 'B', 'C'],
    initial='A', 
    auto_transitions=False,
    transitions=[
        {'trigger': 'e0', 'source': 'A', 'dest': 'B', 'conditions': ['ready']},
        {'trigger': 'e1', 'source': 'B', 'dest': 'C', 'conditions': ['bored'], 'unless': ['bad-c']},
    ],  
    show_conditions=True,
)   

g = fsm.get_graph(title='Condition example', force_new=True)
g.draw('cond-ex.png', prog='dot')
```
cond-ex

transitions/extensions/diagrams.py
```diff
@@ -74,7 +74,23 @@ def _add_edges(self, events, container):
                 src = transitions[0]
                 for t in transitions[1]:
                     dst = t.dest
-                    container.add_edge(src, dst, label=label)
+                    lbl = self._transition_label(label, t)
+                    container.add_edge(src, dst, label=lbl)
+
+    @staticmethod
+    def edge_label_from_transition_label(label):
```
tyarkoni: Is there a reason to have this as a static method in this class? I don't see it invoked from anywhere other than the tests right now--in which case it should probably be moved into the test class.

tyarkoni:
Looks good to me, except I would suggest maybe adding a space before and after the ampersand (just for readability).

aleneum:
> Here is code example and result in attached png file
:+1: 
> except I would suggest maybe adding a space before and after the ampersand
:+1:

transitions/extensions/diagrams.py
khigia: Moved it.

Reasoning to leave it in AGraph was to get it close to reverse code `_transition_label` as both are making same assumption on formatting. But it doesn't matter really.

khigia:
Space added around ampersand.

Issue #89, opened 2016-03-08
Logger as parameter to use with multiple machines


khigia:
Use of standard logging is great. However there are use cases where more than one FSM can be instanciated, in which case it is difficult to match log lines with code logic.

I would like to propose an optional logger parameter when instanciating the machine (with default to current global logger). This way user has possibility to pass a logger instance embedding some context and configuration.

wtgee:
Seems like you could just pass some kind of `name` or `id` attribute of the FSM to the logger rather than passing multiple loggers around. 

I'm not entirely sure this is something that most users would need and that we would want to incorporate but will wait for other opinions.

khigia:
IMHO using name or id is not easier; passing a logger allow user to embed any context (using LoggerAdapter) without having FSM code to have more data (name or id); and logger has advantage of having its own logging conf (level etc). It's completely transparent and optional.

Btw, if this feature is seen as useful (or at least not creating issue I haven't foreseen), I can maybe propose a pull request (to be honest I haven't looked at all extensions code though).

And one last question: is creating issue a good way to propose feature? let me know and I will use the project standard workflow!

Thanks you very much.

tyarkoni:
I think I'd be okay adding an optional logger argument to the Machine constructor. It seems like a pretty minimal change, and I can see a use case for it. If you want to submit a PR, that would be great.

And yes, feature requests should go here.

khigia:
Pull request created: https://github.com/tyarkoni/transitions/pull/91

khigia:
Hum, in core, having a logger as attribute is nice. But in extensions (because of deepcopy and pickle use) the code is a bit convoluted. Maybe better to close the pull request and add a name attribute to Machine as @wtgee proposed. Let me know.

aleneum:
I guess your implementation is pretty close to the minimal changes necessary to make this work.
So `__setstate__`, `__getstate__` and `_deepcopy_` have to be written and/or altered. This is quite some overhead imho. However, deepcopy should not be necessary any longer if `dev-nested` gets merged.  Being able to deep copy machines is nice though if you ask me.
Concerning this I favour the 'machine id/name' solution and -- optionally --  extending the logging string a bit.

aleneum:
I have added an identifier string to Machine in `dev-0.4`.

machine(initial='A', name='Test Machine') results in:
machine.to_B()
machine.to_C()
```
-------------------- >> begin captured logging << --------------------
[...]

PR #91, opened 2016-03-10
Support specific logger per machine.

khigia:


examples/logger_per_machine.py
```diff
@@ -0,0 +1,35 @@
+"""
+When instanciating multiple Machine, it can be useful to set different
```
Ch00k: Typo - instantiating.

aleneum:
I close this since #89 has been implemented in `dev-0.4.0` based on the `name` keyword. If you think that's not sufficient, feel free to reopen again.

Issue #90, opened 2016-03-08
Triggering event inside callback


khigia:
It looks like we can trigger an event from inside a callback (e.g. on_enter_mystate). But this is probably a source of confusion: the callback code start executing while the FSM is in one state, call the event (FSM might change state and trigger other callbacks), and first callback finish executing in a different FSM state (we can see strange nested on_entering, on_exiting log).

I believe a less surprising behaviour would be to queue event execution while a transition is already executing. Would that make sense?

wtgee:
Do you have some example code that demonstrates the problem and/or solution? Are you using the threaded version of the FSM or just a regular?

khigia:
I am using the regular version (I believe the locked version would block).

Here is an example of code triggering this behaviour:
`import logging
import transitions
transitions.logger.setLevel(logging.INFO)
logging.basicConfig()

class Model(object):
    
    def on_enter_B(self):
        print('begin enter B', self.state)
        self.e1()
        print('end enter B', self.state)


m = Model()

fsm = transitions.Machine(
    model=m,
    states=['A', 'B', 'C'],
    initial='A',
    transitions=[
        {'trigger': 'e0', 'source': 'A', 'dest': 'B'},
        {'trigger': 'e1', 'source': 'B', 'dest': 'C'},
    ],
)

m.e0()
`

And here is the output:
`INFO:transitions.core:Initiating transition from state A to state B...
INFO:transitions.core:Exited state A
('begin enter B', 'B')
INFO:transitions.core:Initiating transition from state B to state C...
INFO:transitions.core:Exited state B
INFO:transitions.core:Entered state C
('end enter B', 'C')
INFO:transitions.core:Entered state B
`

As we can see, state change during execution of `on_enter_B`. And logging become rather confusing as we log 'Exited B' before 'Entered B', and after this log line, state is C. Having more callback (before/after) transition would probably be difficult to handle.

I have not implemented a solution but I believe that queueing call to `transitions/core.py` trigger method would solve the issue (by queuing I mean delaying the processing, this can be done on same thread).

tyarkoni:
[...]

PR #92, opened 2016-03-11
Queue trigger to allow event in callback

khigia:
This branch allow queuing event trigger in callback as discussed in issue #90.


aleneum:
please issue again for `dev-async` branch

PR #95, opened 2016-03-20
Return immediately if transition not found while ignore_invalid_triggers is True

Ch00k:


aleneum:
Coverage decreased? [Details](https://coveralls.io/builds/5479865)  look good! Thank you @Ch00k!

Issue #86, opened 2016-02-13
Improve nested extension


aleneum:
Since my first implementation I got some feedback about the way I have done Hierarchical State Machines in transitions. The current implementation has some issues:
* **underscore separator does not cut it**: We have been aware of the fact, that underscore is no perfect solution but it allowed nested states to reuse many of transitions' core functionality (auto transitions and callback addition). However, since underscore can also be used in state naming, there is no way to tell if `stateC_alt_state1` is actually a (two times) nested state or not. Even worse could be a situation where the naming of states collide (`stateC` + `alt_state1`; `stateC_alt` + `state_1`). I haven't even tried what the ambiguity in auto transition would do but `to_stateC_alt_state1` would probably do random things.
* **flatten strategy destroys hierarchical information**: After flattening, retrieving hierarchical information, especially related to transitions, is bothersome. You cannot rely on the name. This means you have to retrieve the state in question and have a look at the parent (and its parent and so on) to actually know HOW nested the state was. I still think flattening is not a bad approach but especially in combination with the underscore separator it makes things unnecessarily complicated. The blueprint strategy could compensate this a bit and made copying possible but its one of the things that is hard to comprehend and maintain especially if the 'causer' is not available anymore.

I spent some time thinking about it and came to the conclusion that I have to change some things under the hood but also violate the first law of a library developer which is *'One shall not break the API'* by **switching from underscore to a dot separator**. Why choosing the dot:

* It's already implicitly forbidden in naming because it makes the convenience features of transitions useless: `'my.state' -> to_my.state (Nope) -> on_enter_my.state (Nope)`
* From python's point of view it's suitable to imply nesting: `machine.to_stateA.alt.state1` resembles common `object.child.property` nesting. I have chosen `to_stateC.alt` rather than `to.stateC.alt` to stick to the way transition's core work. This is done with a callable [FunctionWrapper](https://github.com/tyarkoni/transitions/compare/dev-nested#diff-0340de4ea0a66c6a0af24ea2b01ea855R11) which contains auto transitions to child states as properties (wrapping happens [here](https://github.com/tyarkoni/transitions/compare/dev-nested#diff-0340de4ea0a66c6a0af24ea2b01ea855R231)).

[...]

PR #96, opened 2016-04-12
Push to 0.4.0

aleneum:
I have changed quite a lot in this dev-branch but the impact on `core` has been kept low.
The addition of the feature requests #89 and #90 went into core. #86 is mostly limited to `nesting`.
Some changes in `diagram` have been necessary to allow the highlighting of clusters (parent states) in the dot diagram. I have also merged `AAGraph` into `AGraph` since it was exclusively used even in non-nested scenarios.

Bugfixes include:
* updating `graph` in `diagram` if a state or transition is added; otherwise every triggered event will fail since the dot node/cluster is not known in `graph`
* Mixin of `LockedHierarchicalGraphMachine` was wrong which caused the wrong `create_event` method to be called
* updated setup.py with `find_packages` to actually package extensions into the build (have been omitted so far)
* Lots of testing code

I also extended the documentation in `ReadMe` with the (previously undocumented) features and also added more hyperlinks to the table of content.

Of course, the version push to 0.4.0 is just a suggestion since a lot of stuff has changed under the hood and new features have been added. If you have questions or feedback or feel like something needs more code comments I am glad to assist.


tyarkoni:
Yes, definitely--this is long overdue. I'm on vacation till Saturday, but will push the new release on Sunday. Thanks!

aleneum:
>  coverage/coveralls — First build on dev-0.4 at 98.324%

travis seems to be a bit impatient :)

PR #106, opened 2016-05-25
Tox

aisbaa:
Hi, I've added support for [tox](https://tox.readthedocs.io/en/latest/). Should help testing transitions with many interpreters on local machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ dist/*
 .idea
 .coverage
 .ipynb_checkpoints
+.tox/
 examples/state.png
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - "3.5"
   - "3.4"
   - "3.3"
   - "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ python:
   - "3.3"
   - "2.7"
 install:
-  - pip install python-coveralls coverage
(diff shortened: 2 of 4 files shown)


tyarkoni:
Thanks for the PR, @aisbaa. Could you provide a quick summary of the concrete benefits? I'm not familiar with tox, and will look at the docs when I have a chance, but I'm kind of swamped at the moment. It's not immediately obvious to me what we gain over the current setup. (Just to be clear, I'm not expressing any skepticism about the benefits, I just don't have the time to look into it right now. @aleneum or @wtgee may have a more informed opinion.)

aisbaa:
Will try to do my best. Basically tox tries to minimize differences between all testing environments by providing really strict isolation from system.

1. It creates separate virtualenv for each interpreter (missing interpreter can be skipped). By running tox you would be running tests on all available python interpreters.
2. Virtualenv does not include system libraries (unless configured to include) - therefore there is less of a chance that tests run on someones computer and don't run on other because of dependency misconfiguration.
3. Before running tests it installs latest version of your library into virtualenv, so setup.py gets tested.

Hope this helps!

aleneum:
That does sound interesting and I already have a project in mind where I definitely will try it out :).
Regarding transitions/github I'd say that most of the functionality tox offers is already covered with travis. I might be wrong though.

aisbaa:
@aleneum yes indeed most is already covered by travis, will close this pull request ;)

PR #107, opened 2016-05-31
Fix #105 issue - missing edges between nodes

### LINKED ISSUE #105: GraphMachine does not include all edges in dot file
I think I found a bug. Code below should produce graph with 4 directional edges, but it gets only one.

```python

class AToB(object):
    pass

transitions = [
    ['event_0', 'a', 'b'],
    ['event_1', 'a', 'b'],
    ['event_2', 'a', 'b'],
    ['event_3', 'a', 'b'],
]

machine = Machine(
    AToB(),
    states=['a', 'b'],
    transitions=transitions,
    initial='a',
    auto_transitions=False,
)

machine.graph.draw('/tmp/missing_edges.png', prog='dot')
```

transitions==0.4.0

aisbaa:
Simple fix for missing edges in graph machine. Let me know if different approach is also desired to this problem.

### DIFF
--- a/tests/test_graphing.py
+++ b/tests/test_graphing.py
@@ -149,3 +149,29 @@ class TestDiagrams(TestCase):
         m.add_state('X')
         m.add_transition('foo', '*', 'X')
         m.foo()
+
+    def test_if_multiple_edges_are_supported(self):
+        transitions = [
+            ['event_0', 'a', 'b'],
+            ['event_1', 'a', 'b'],
+            ['event_2', 'a', 'b'],
+            ['event_3', 'a', 'b'],
+        ]
+
+        machine_cls = MachineFactory.get_predefined(graph=True)
+        m = machine_cls(
+            states=['a', 'b'],
+            transitions=transitions,
+            initial='a',
+            auto_transitions=False,
+        )
+
+        graph = m.get_graph()
+        self.assertIsNotNone(graph)
+        self.assertTrue("digraph" in str(graph))
+
+        print(str(graph))
+
+        triggers = [transition[0] for transition in transitions]
+        for trigger in triggers:
+            self.assertTrue(trigger in str(graph))
--- a/transitions/extensions/diagrams.py
+++ b/transitions/extensions/diagrams.py
@@ -108,8 +108,7 @@ class AGraph(Diagram):
                         while len(dst.children) > 0:
                             dst = src.children
 
-                    if container.has_edge(src.name, dst.name) is False:
-                        container.add_edge(src.name, dst.name, label=label, ltail=ltail, lhead=lhead)
+                    container.add_edge(src.name, dst.name, label=label, ltail=ltail, lhead=lhead)
 
     def _transition_label(self, edge_label, tran):
         if self.machine.show_conditions and tran.conditions:


aleneum:
Unfortunately it is not that easy. The check you want to remove is required for nested graphs where the parent has a valid transition to it's first child and the child itself also has such a transition (for instance because of '*'->parent.child1). Removing it will cause duplicated edges for all *-transitions in nested graphs. There has not been a test for that yet, sorry.

However, I do think the current 'duplicate fix' is not sufficient at all if it prevents having multiple edges with different names. It also 'hides' the edge from parent to child. This is a limitation in graphviz/dot (or how it deals with edges from cluster to nodes) I could not work around yet.

Long story short: I think we should merge the pull request to allow multiple edges and I will open  a new issue to find a better to solution to deal with cluster-to-node-edges.

tyarkoni:
Okay, but in view of the nesting issue, maybe we should just go with my suggestion in #105 to only update the labels rather than allowing multiple edges between the same vertices. Is there a downside to that approach? I'm not sure we need to allow multigraphs in this context. Any FSM with a reasonable number of states is already going to look very messy as is...

wtgee:
I'm with @tyarkoni as I think the labels idea is a little better although that still gets a little crowded on any reasonably sized FSM.

aleneum:
Clutter is a good argument. How should it be annotated then? Like

`event[condition&condition&!unless] || event2[condition&!unless&!unless]`  ?

aleneum:
Example:

```python
states = ['standing', 'walking', {'name': 'caffeinated', 'children':['dithering', 'running']}]
transitions = [
  ['walk', 'standing', 'walking'],
  ['go', 'standing', 'walking'],
  ['stop', 'walking', 'standing'],
  {'trigger': 'drink', 'source': '*', 'dest': 'caffeinated_dithering',
   'conditions':'is_hot', 'unless': 'is_too_hot'},
  ['walk', 'caffeinated_dithering', 'caffeinated_running'],
  ['relax', 'caffeinated', 'standing']
]

machine = GraphMachine(model=Matter(), 
                       states=states, 
                       transitions=transitions, 
                       auto_transitions=False, 
                       initial='standing', 
                       title="Mood Matrix",
                       show_conditions=True)
```
leads to

state

wtgee:
I just tested on my main machine which is modest (I think?) at 11 states. Each state has one `check_safety` condition and some of the states have an additional condition. The generated graph is pretty much too large to use. But I guess if you printed it and put it in a wall for reference it would be handy. :)

I expect not too many people will use this on large machines but it's definitely a feature we should support.

wtgee:
After looking at my large graph closer, either there is a bug in the code or this will be really helpful in finding conditions that are applied multiple times.

If you look at the `park` transition on the top it has the `check_safety` condition listed multiple times and I'm pretty sure my code isn't doing that. I'll look into it more and update shortly.

State File

aleneum:
yeah, I fixed that but wanted to wait for your oppionion about 

> event[condition&condition&!unless] || event2[condition&!unless&!unless]

If you are okay with it I would pull @aisbaa test and commit my changes to make it work with nested graphs

wtgee:
@aleneum Got it, sorry. I'm with it now. Is your fix small that you can post a gist here so I can see run it on my actual machine?

aleneum:
The relevant part is just two lines in extenstion/diagram.py

https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6

problem is, that `label` is falsely reused.

[this](https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6#file-fix105-diff-L10) and [this](https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6#file-fix105-diff-L29) should be enough.

wtgee:
Great, thanks. Looks a lot better for sure. :+1:

On Wed, Jun 1, 2016 at 6:48 AM, Alexander Neumann <notifications@github.com>
wrote:

> The relevant part is just two lines in extenstion/diagram.py
>
> https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6
>
> problem is, that label is falsely reused.
>
> this
> <https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6#file-fix105-diff-L10>
> and this
> <https://gist.github.com/aleneum/602def83e94bf4ee38907d8151cef9b6#file-fix105-diff-L29>
> should be enough.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/tyarkoni/transitions/pull/107#issuecomment-223054548>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAEUUI5DZATKO9RKEVUvKiTfJfSPI7Moks5qHbfygaJpZM4IqgYc>
> .
>



-- 
~Wilfred Tyler Gee

PR #108, opened 2016-06-02
Async rename issues 101

### LINKED ISSUE #101: async naming
I've been using and testing out the `async` feature for queued processing and all seems to be working well however I find myself wondering at the name `async`. Wouldn't it make more sense to have the named parameter be `queued` or something? 

Not a big deal, but this feature doesn't really have anything to do with the `async` module and I think it might lead to some confusion for people who are trying to use some of python's new `async` features.

Thoughts? (Also, sorry for not catching this before it went into master but I don't think we've done an actual release yet...)

wtgee:
Async Rename:
* `has_queue` property for predicate
* closes #101 

Misc:
* Arrange import statements.
* Small cleanups


aleneum:
👍

aleneum:
Minor spelling error [here](https://github.com/tyarkoni/transitions/pull/108/files#diff-ead0bb6c6c63bccefe005346890f03a5R311)

is `queed` should be `queued`.

PR #109, opened 2016-06-16
Fix 'RuntimeError: maximum recursion depth exceeded'

dcere:
When using the word `process` as a trigger a `RuntimeError` exception is raised due to the `process` method calling itself infinitely. Changing the name of the method to `_process` fixes that.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -532,3 +532,10 @@ class TestTransitions(TestCase):
         # An invalid transition shouldn't execute the callback
         with self.assertRaises(MachineError):
                 m.model.on_exit_A()
+
+    def test_process_trigger(self):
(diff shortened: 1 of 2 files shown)


aleneum:
You are right. This should be a 'protected/private' method and also it was just a matter of time until somebody stumbles upon this issue when using the trigger name 'process'. Thank you for your contribution!

Best regards,
Alex

wtgee:
Thanks for catching that @dcere. Similarly, one could hit the same error with a trigger named `callback`:

```python
def foo(self): print("FOO!")

transitions = [
    { 'trigger': 'callback', 'source': 'before', 'dest': 'after', 'before': 'foo'},
]

machine = Machine(states=['before', 'after'], transitions=transitions, initial='before')
machine.callback()
```

Any reason to keep the `callback` method public?

Issue #112, opened 2016-07-01
__init__ methods don't call super


jonathanunderwood:
I've been trying to use transitions in a project where I want classes to inherit from Machine (or one of the extended variants) as well as other objects, and unfortunately Machine doesn't play nice - it doesn't accept *args, **kwargs, and doesn't call `super().__init__(*args **kwargs)` (py3), or `super(Machine,self).__init__(*args, **kwargs)` (py2/py3). That means that the usual expectations of multiple inheritance and the mro aren't really honored. Any chance of fixing this?

Otherwise, I am loving this package, thanks so much for making it available.

aleneum:
Yeah,

this happens when Machine is not the 'last in line'. I see where Machine calling super could be useful. To estimate the changes required it would be nice to know if you also plan to inherit from `State`, `Transition` or `Event`.

aleneum:
Additionally,

could you provide a minimal example of your inheritance scenario to test against.

jonathanunderwood:
In my work to date, I haven't needed to inherit from State, Transition, or Event - I've only inherited from Machine and HirearchicalStateMachine - I can imagine inheriting from the other extended Machines too.

The issue is actually two things: calling super, and also handling args and keyword args in a way that init functions through the MRO can extract their parameters as needed. 

Here's a simple example for the Machine class. Notice that when creating the D instance, neither A or B's init functions are called:

```python
from transitions import Machine

class A(object):
    def __init__(self, *args, **kwargs):
        print('A init')
        super(A, self).__init__(*args, **kwargs)

class B(object):
    def __init__(self, *args, **kwargs):
        print('B init')
        super(B, self).__init__(*args, **kwargs)

class C(A, B):
    def __init__(self, *args, **kwargs):
        print('C init')
        super(C, self).__init__(*args, **kwargs)

class D(Machine, A, B):
    def __init__(self, *args, **kwargs):
        print('D init')
        super(D, self).__init__(*args, **kwargs)


print('Making a C instance:')
c = C()

print('Making a D instance:')
d = D()
```

jonathanunderwood:
Here's a slightly more complicated example showing how arg and keyword arg handling should work...

```python
from transitions import Machine

class A(object):
    def __init__(self, aarg, *args, **kwargs):
        print('A init, aarg:{0}'.format(aarg))
        super(A, self).__init__(*args, **kwargs)

class B(object):
    def __init__(self, barg, *args, **kwargs):
[...]

Issue #113, opened 2016-07-01
Semantics of HierarchicalStateMachine


jonathanunderwood:
Using the example given in the README for the heriarchical state machine variant, if one executes

```python
machine.to_caffeinated_dithering()
```
then I think it would make sense for this to return `True`:

```python
machine.is_caffeinated()
```
whereas at present it returns `False` since caffeinated_dithering is a sub-state of caffeinated (of course, `machine.is_caffeinated_dithering()` returns `True`).

I realize this would break backwards compatibility, so perhaps such a change would be activated by supplying a keywrod argument to the __init__ function of HSM.

aleneum:
Hi @jonathanunderwood,

nice idea. What about this: is_state could accept a `strict` keyword which allows substates:

```python
machine.state # caffeinated_dithering
machine.is_caffeinated() # returns True
machine.is_caffeinated(strict=True) # returns False
```

Question is, if strict should be default true or false.

jonathanunderwood:
Yes, that is a nice interface design, I agree. "specific" might be a better keyword than "strict", but that's just bike shedding on my part.

If you don't want to break people's existing code, I think you'd want strict to default to True.

jonathanunderwood:
Wow, that was quick work, many thanks!

aleneum:
see merge

PR #114, opened 2016-07-01
Fixes Issues #112 and #113

aleneum:
Added a super call in `Machine` (#112) and added optional `allow_substates` argument for `HierarchicalStateMachine.is_state` (#113).

The super call is surrounded with try/except to prevent cryptic `object.__init__() takes no parameters` errors if wrong parameters are passed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.4.2 (... 2016)
+Release 0.4.2 contain minor API changes:
+- HierarchicalStateMachine.is_state now provides `allow_substates` as an optional argument(thanks to @jonathanunderwood)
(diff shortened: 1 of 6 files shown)


aleneum:
Since this pull request will change some things in the core I will wait for some feedback from @wtgee or @tyarkoni. It is supposed to not break anything. [This](https://github.com/tyarkoni/transitions/pull/114/files#diff-ead0bb6c6c63bccefe005346890f03a5R316) is the most relevant change. If someone passes wrong parameters to a Machine like `befor` instead of `before`, this now raises a `MachineError`. Previously issues like that had been ignored. It also allows more flexible handling of multiple inheritance (as discussed in #112 ). This might also be relevant for all the guys that want to mix transitions into django apps. What do you think?

wtgee:
Sorry, took me a while to look a but looks good. :+1:

PR #127, opened 2016-08-07
Fix link in README

otommod:
Even though the link for the Transitions "chapter" was explicitly set to ```transitions```, markdown instead created that link for the name of the module.

So set the link to the module name to ```transitions-module``` and fix the problem.


tyarkoni:
Thanks!

PR #135, opened 2016-08-28
Multiple models

aleneum:
Since all tests pass and the example notebooks also look good I propose a merge into `master`.
What do you think?


aleneum:
I will commit improved test coverage tomorrow.

tyarkoni:
+1 for merging. More test coverage would be nice, but since all the single-model tests seem to pass now, and probably 99% of use cases will continue to be with one model, I don't know that we need to invest a lot of energy writing multiple-model versions of all the tests.

aleneum:
Since we just talk about some lines, I wanted to give it a try. This was also a nice opportunity to find unreachable code which could be removed/simplified. I found two occasions:

```python
def model(self):
        ...
        elif not self.models:
            return None
       ...
```
self.models is set during initialisation and only tinkering with internals of `Machine` could lead to triggering this. In this case it might be better to let the standard `AttributeError` inform the user of the issue instead silently returning None. This could be removed in my opinion.

Second (and the harder one):

```python
    def __getattr__(self, name):
        if name.startswith('__'):
            if name in self.__dict__:
                return self.__dict__[name]
            else:
                raise AttributeError("{} does not exist".format(name))
```

I cannot think of an example where dict contains keys starting with double underscore. `Machine`'s dict does not contain any methods like `__getattr__` and if you try to add double underscore methods (to the class for instance), they will be mangled as far as I know. I am sure it is possible somehow but does it really 

This:
```python
if name.startswith('__'):
    raise AttributeError("{} does not exist".format(name))
```
should be sufficient until somebody can prove that having double underscore keys is required for some use cases.

PR #136, opened 2016-09-05
add attribute naming checks to prevent overrides of model.trigger

aleneum:
I added a check whether a model already contains a ‘trigger’ attribute to prevent accidental overriding. 

I am a bit unsure about what a decent ‘plan B’ would be.
My initial idea: Adding ` _trigger` instead.
This would allow a model to change the behaviour of the actual trigger method:

e.g.

```python
class Model:
    def trigger(self, *args, **kwargs):
        if not self._trigger(*args, **kwargs):
            # do something if transition fails
```

This makes default transitions as mentioned by @wtgee easier.
But maybe you have a better idea of how to handle naming collisions in this case.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -621,11 +621,7 @@ class TestTransitions(TestCase):
 
         model = Model()
         m = Machine(model=model)
-        self.assertTrue(hasattr(model, '_trigger'))
         self.assertEqual(model.trigger(5), 5)
-        model._trigger = lambda x: x
-        m = Machine(model=model)
-        self.assertEqual(model._trigger(5), 5)
 
 
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -605,6 +605,10 @@ class TestTransitions(TestCase):
         def return_value(value):
             return value
 
+        class Model:
+            def trigger(self):
+                pass
+
         self.stuff.machine.add_transition('do', '*', 'C')
         self.stuff.trigger('do')
         self.assertTrue(self.stuff.is_C())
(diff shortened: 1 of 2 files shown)


wtgee:
I think I would almost prefer a more descriptive word rather than a semi-private one. Maybe something like `do_trigger` or `call_trigger` instead of `_trigger`, which I think would be more inline with some of our other `<verb>_<object>` methods.

tyarkoni:
I don't know that we need the fallback. I doubt this will be functionality that someone expects to exist (and even if they do, they won't necessarily expect it to exist at `.trigger()`; if someone discovers it, it'll be via the documentation. And it seems very unlikely that someone would (a) already be using `.trigger` and (b) be unable to change it if they want to take advantage of this feature. So I would lean towards just skipping binding with an info message if `.trigger` already exists.

tyarkoni:
Alternatively, we could allow the user to name the `.trigger` method themselves. But that seems like overkill to me given that this is likely to be a pretty uncommon use case.

aleneum:
> I think I would almost prefer a more descriptive word rather than a semi-private one. 

I see your point and agree. A more descriptive name would be favourable.

>  I doubt this will be functionality that someone expects to exist [...] So I would lean towards just skipping binding with an info message if .trigger already exists.

I think you are right. The approach is kind of a hack. I'd like to have the opportunity to extend `Model.trigger` somehow to specify the behaviour (like having a default transition if `Model.trigger` returns False). However, I do not know how to 'inherit' from functions which are added at runtime. But I assume that if someone requires that behaviour, it's better if he or she explicitly tinkers around with the model rather than `transitions` trying to read minds.

Thank you for your feedback!

tyarkoni:
Looks good to me!

PR #137, opened 2016-09-06
Allow guarding ordered transitions with conditions

limdauto:


transitions/core.py
```diff
@@ -543,14 +543,20 @@ def add_ordered_transitions(self, states=None, trigger='next_state',
...
                 placeholder to be included in the added transitions.
+            conditions (None or list): A optional list of conditions that guard
+                in front of each corresponding ordered transition
         """
         if states is None:
             states = list(self.states.keys())  # need to listify for Python3
         if len(states) < 2:
             raise MachineError("Can't create ordered transitions on a Machine "
                                "with fewer than 2 states.")
+        if conditions is not None and len(conditions) != len(states):
```
aleneum: Shouldn't it be `len(conditions) != len(states) - 1` since `n` states mean `n-1` transitions.
e.g: A, B, C, D -> A>B, B>C, C>D
Additionally, I'd suggest to also accept string or a function as valid for `conditions` in cases where one condition check should be applied to all transitions
e.g: `is_ready` or similar

PR #143, opened 2016-09-17
log, FIX #139: Kept only enter/exit in log-INFO

### LINKED ISSUE #139: Logger reports everything on >= INFO, no debug-level used
Excellent library, thank you, and the logging statements are crucially placed.

May I suggest that some logging-statements report on a lower level, e.g. those reporting the execution of some callback, and possibly those reporting pass/fail of failed conditions?
(basically, just keep the Enter/Exit log-statements in info level)

Should I provide a PR?

ankostis:
+ Fixed past-tense typo in the log-msgs about executed callbacks.
+ Replace Leaving-->Exiting to match Enter/Entering log-msgs.
+ Log WARN if modes already defines `trigger` method.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -155,7 +155,7 @@ class Transition(object):
             successfully executed (True if successful, False if not).
         """
         logger.debug("%sInitiating transition from state %s to state %s...",
-                    event_data.machine.id, self.source, self.dest)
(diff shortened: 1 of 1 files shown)


wtgee:
Eek, sorry, the new github review thing is messing with me. Still need to get lines 158 and 168 changed. Thanks!

PR #145, opened 2016-09-23
this fixes #140

### LINKED ISSUE #140: Is it possible to add a method to get all eligible triggers from a current state?
I'm thinking of keeping a simple dictionary keyed by state name with a `defaultdict(list)` of transition names that can be executed from this state. I am wondering if this is a responsibility of the core api or should this be an extension?

aleneum:
see #140


theY4Kman:
Great job there, coveralls. Really staying on top of things.

Issue #128, opened 2016-08-10
same state machine instance for multiple object instances


gemerden:
In our use case we have many objects that could be handled with the same state machine (e.g. from database queries). Currently a state machine is instantiated for each object which creates (perhaps unnecessary) overhead.

Would it be possible to have a version of the state machine that points self.machine in multiple objects (of the same class) to a single state machine instance?

tyarkoni:
This is an interesting idea. In theory there's no principled reason why you shouldn't be able to use the same `Machine` for multiple models, provided you want identical behavior. Right now the only barrier I see to doing that is that we track the current `State` object within the `Machine` instance  rather than the model instance. This is actually a design flaw, in my opinion; I think it would be cleaner and more sensible to maintain state only in the model. As it stands we already track the current state *name* within the model, but not the `State` instance itself.

This should be easy to fix without breaking the API. Assuming we move all state out of the `Machine`, we could then modify the initializer to take a list of models rather than just a single model. Any changes applied to the machine (e.g., adding new transitions) would thereafter be applied to all available models. So that would work fine in cases where you have all of your models instantiated in advance of `Machine` creation. But note that you wouldn't be able to easily add extra models *after* initializing the `Machine`. Simply pointing to the right `Machine` in your models wouldn't give you the desired behavior, since you wouldn't automatically have all the dynamically added triggers and callbacks. For that, we'd need to implement some kind of `add_model()` method in `Machine` that re-applies all existing transitions and callbacks. That's also potentially doable, but I'm not sure I want to take that step unless this is going to be a fairly common use case (whereas simply managing multiple existing models should be trivial).

Anyway, thanks for opening this issue. I'll think about it some more, and we should be able to do at least something to accommodate this use case without having to change much.

gemerden:
Thank you for your quick reply. I think having just the state name in the model is not a problem; it is always possible to access the actual state from the model and from a database persistence point of view having just strings for the state is much easier to handle. 

After reading some more of the transitions code i think it would be possible to maintain (just extend) the API, and optionally move all methods (callback, triggers) to the state machine itself, by introducing an optional base class for the models. E.g. for triggers something like this:


```
class BaseModel(object):

    machine = Machine(states, transitions)  # e.g. have one state machine instance for all instances 

    def __getattr__(self, trigger_name):  # only called if  trigger_name was not found on the model
        if trigger_name in self.machine.triggers:  # or some such, maybe not even required
[...]

PR #148, opened 2016-10-06
Fix GraphMachine usage Example

ketanbhatt:
The attribute `graph` is not set on the instance of the model, but on the machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,7 +803,7 @@ Transitions can generate basic state diagrams displaying all valid transitions b
 from transitions.extensions import GraphMachine as Machine
 m = Model()
 machine = Machine(model=m, ...)
-m.graph.draw('my_state_diagram.png', prog='dot')
+machine.graph.draw('my_state_diagram.png', prog='dot')
 ```
 
 This produces something like this:


wtgee:
HI @ketanbhatt. Since we moved to support multiple models (#128) the graph actually became a property of an individual model rather than the machine. So the example is actually correct, however I can easily understand where the confusion comes from. A `GraphMachine` does have a `get_graph` method which is meant to be used to get that graph so maybe the example would read better as:

```python
m = Model()
machine = Machine(model=m, ...)
machine.get_graph().draw('my_state_diagram.png', prog='dot')
```

In the [example notebook](https://github.com/tyarkoni/transitions/blob/master/examples/Graph%20MIxin%20Demo.ipynb) you can also see how we define a convenience method on the model called `show_graph` with the point being that it is actually a property of an individual model.

However if all models should always be in the same state this might actually not be that intuitive. @aleneum @tyarkoni it almost seems to me that it makes sense to move the graph back into the machine rather than the model. The graph, after all, reflects the current status of the machine, not of any individual model. Or is there something I'm missing with how the multiple models should be used?

tyarkoni:
@wtgee, models can now be in different states, so I think it makes sense to keep the graphs in the model. See discussion in #128.

aleneum:
Hi @wtgee,

> However if all models should always be in the same state this might actually not be that intuitive.

that is not necessarily the case. Every model can be in a different state. This [test](https://github.com/tyarkoni/transitions/blob/master/tests/test_core.py#L586) in `test_core` shows how it works (I adapted it to be standalone)

```python

class Model:
    pass

def test_multiple_models():
    s1, s2 = Model(), Model() 
    states = ['A', 'B', 'C']

    m = Machine(model=[s1, s2], states=states,
                initial=states[0])
    #self.assertEquals(len(m.models), 2)
    #self.assertEquals(len(m.model), 2)
    m.add_transition('advance', 'A', 'B')
    s1.advance()
    #self.assertEquals(s1.state, 'B')
    print s1.state # >>> B
    #self.assertEquals(s2.state, 'A')
    print s2.state # >>> A

test_multiple_models()
```

ketanbhatt:
@wtgee so I tested it out, and if I do:

```python
m = Model()
machine = Machine(model=m, ...)
m.graph.draw('my_state_diagram.png', prog='dot')
```

I get: `AttributeError: 'Model' object has no attribute 'graph'`

It works if I do `machine.graph.draw('my_state_diagram.png', prog='dot')`. Can you check if that is the case with you too?

wtgee:
@tyarkoni @aleneum right, sorry, now my brain is refreshed on that. I haven't yet worked with it in multiple models and August was the month I was much too busy with other things.

@ketanbhatt You are most likely on a slightly outdated version from what the documentation reflects.

ketanbhatt:
Yes that does seem to be the case. I am on `0.4.1`, and I installed using `pip`.

But `transitions/version.py` also says that is the version, and that seems to be the highest number according to tags too. Am I missing something?

aleneum:
Hi @ketanbhatt,

the tag [0.4.1](https://github.com/tyarkoni/transitions/tree/0.4.1) happened about 30 commits ago. We are currently working on release 0.4.2 . If you like to use multiple models as well as the other new features mentioned in the [Changelog](https://github.com/tyarkoni/transitions/blob/master/Changelog.md) you have to install transitions manually at the moment. If you are okay with 0.4.1, have a look at the [documentation](https://github.com/tyarkoni/transitions/blob/0.4.1/README.md) for the tag.

ketanbhatt:
Okay, thank you. I will close this then.
Also, the library is written very beautifully, we are going to use it in our codebase, and it fits in perfectly. 

Thank you again.

tyarkoni:
Sorry, 0.4.2 is long overdue. Will try to push to PyPI today.

ketanbhatt:
Will update in that case 😄

tyarkoni:
Done!

Issue #147, opened 2016-09-25
more diagram modes


aleneum:
I want to propose two new modes/functions for the graphing extension:
- `keyword: filter_auto_transitions` -- having auto transitions enabled causes clutter if state machines become more complex. Imho, they do not add information to the graph because it should be quite obvious how every 'to_'-method works. At the moment I disable auto transitions for 'productive' deployment but I would prefer to have a useful graph AND auto transitions.
- `keyword: roi_only/show_roi_only` -- having a 'focused' view of the graph is another feature that may be useful. Since the state machine I am currently working with has about 15 states (and growing) and most of the time only a fraction of these states is relevant. A focused view would include the last state, (the last successful transition,) the current state and all possible transitions from this state.

The keywords could be part of the constructor or passed to the `get_graph` function. Especially `roi_only` might be something you want to use in conjunction with the 'standard' behaviour and therefore in `get_graph`.

What do you think about this? Any suggestions for the naming of these parameters?

wtgee:
:+1: I think it sounds like a good idea and I could see both of them being useful.

Does it make sense to automatically filter the auto transitions and then make the keyword `display_auto_transitions`? It might lead to some confusion if people don't read the docs carefully enough.

(Also, sorry for ignoring this for 17 days...)

tyarkoni:
Sorry, I also missed this. +1 on both ideas as well. And also +1 on @wtgee's suggestion to not display auto_transitions by default, and call the flag `display_auto_transitions` (or maybe `show_auto_transitions`).

aleneum:
* added `show_roi` to `get_graph` function which is attached to the model.
* added `show_auto_transitions` (default False) to `GraphMachine.__init__`

both has been integrated into `dev-0.5`

PR #151, opened 2016-10-17
Feature #147

aleneum:
See #147. Have a look at the [example notebooks](https://github.com/tyarkoni/transitions/tree/feature-%23147/examples) to see it in action.

I went for `show_auto_transitions` to be consistent with `show_conditions` which we already have.
I shortened `show_roi_only` to `show_roi` but can undo that if this is considered less comprehensible.


aleneum:
Just fyi:

Coverage decreases because of [this](https://github.com/tyarkoni/transitions/pull/151/files#diff-e29d8b8e9d103e0a5c6a8746a56a0f0cR265) line.
Python optimises some `continue` statements and therefore coverage cannot evaluate it. More info [here](https://bitbucket.org/ned/coveragepy/issues/198/continue-marked-as-not-covered).

README.md
```diff
@@ -793,24 +793,30 @@ machine = Machine(model, states, transitions)
...
-Transitions can generate basic state diagrams displaying all valid transitions between states. To use the graphing functionality, you'll need to have `pygraphviz` installed (`pip install pygraphviz`). With
- `GraphMachine` enabled, a PyGraphviz `AGraph` object is generated during machine initialization and is constantly updated when the machine state changes:
+Transitions can generate basic state diagrams displaying all valid transitions between states. To use the graphing functionality, you'll need to have `pygraphviz` installed (`pip install pygraphviz`). With `GraphMachine` enabled, a PyGraphviz `AGraph` object is generated during machine initialization and is constantly updated when the machine state changes:
 
 ```python
 m = Model()
 machine = Machine(model=m, ...)
-m.graph.draw('my_state_diagram.png', prog='dot')
+# Machine(model=m, show_auto_transitions=True, ...)
```
wtgee: Did you mean to leave this line in here to show the available options? Would be clearer with a comment.

transitions/extensions/diagrams.py
```diff
@@ -133,6 +135,7 @@ def get_graph(self, title=None):
         """ Generate a DOT graph with pygraphviz, returns an AGraph object
         Args:
             title (string): Optional title for the graph.
+            show_roi (boolean): Show only the active region if a graph
```
wtgee: *of* a graph...

aleneum:
I used the opportunity to rework the generation of nested graphs. But I am not 100 percent satisfied with how that worked out. The main problem is that Graphviz/dot does not support edges to clusters. In the past, this has not been a big issue since states with children always invoked their first child. I could do the same for clusters and adjust the edges with the attributes `lhead` and `ltail` to hide certain parts of edges. However, since states now can be entered without the need to enter a child state, this becomes more tricky.

I used a hidden anchor state at first which represents the cluster and sticked to `ltail` and `lhead`. This does look messy in certain cases (lower left):


hide_anchor

In a best case scenario the edges would start and end at the cluster border. But as you see, this does not happen. A transition from the super state to a child state is just floating around and the transition `relax` to state D is very short and the label looks detached.

Thats why I decided to just show the anchor node and its edges:

show_anchor

I'd prefer a solution where edges start and stop at the cluster's border but I could not find a way to achieve this in a generic manner. If you have some thoughts or feedback about how such a graph could be reorganised, let me know.

wtgee:
Yeah, it's a little clunky with the anchor but looks a lot better than the free floating version of this. I'll have a chance to test this out on my machines in the next few days but looks good from the overview.

aleneum:
Okay,

my next attempt: First, I have split Agraph into `Graph` and `NestedGraph`. This should make both graph generation procedures easier to maintain separately as nested graphs require some specific treatment. I hope this makes `Graph` less 'scary' in cases where someone is trying to understand what it is doing while bug hunting. I also removed the leading 'A' because it may be confused with `pygraphviz.AGraph` otherwise.
Second, I reworked the nested graph layout. The subgraph structure *should* take care of having the anchor black dot at the top of each subgraph. You can get the dot 'source' code with `model.graph.string()`. I copied the output, saved it to a file and tried to work out some decent parameters.
I would not claim that these Graphviz parameters are perfectly adjusted but imho its less cluttered for complex graphs than the previous attempt.
To generate a really awesome looking graph, you probably have to fine tune things anyway.
However, if you find better generic combinations of ranking, spacing and grouping, let me know and I will adapt `NestedGraph`.

Example/notebook graph:
state2

Testing graph:
state

aleneum:
I could also rework that a bit and only include anchors if there is a transition from parent to child. In this case the result might me a bit more compact. @wtgee, @tyarkoni: What do you think?

tyarkoni:
Sorry about the delay. This all looks great to me, and I don't think it's worth the extra effort to make it more compact. I agree that a user who wants something pretty and not just functional should expect to have to customize the graph themselves. Okay to merge as far as I'm concerned.

aleneum:
This feature has been integrated into `dev-0.5`

PR #154, opened 2016-10-24
allows passing State instances instead of names; closes #153

### LINKED ISSUE #153: Add possibility to use State object references when calling add_transition
Right now it seems that it's only possible to add transitions by using strings of the state names but not State object references when calling add_transition.

If support was added it would open up for code completion in editors and easily finding misspelled state references since they will be found by static code analysis/pylint.

```
>>> from transitions import State, Machine
solid = State('solid')
liquid = State('liquid')
machine = Machine(states=[solid, liquid], initial=solid)
machine.add_transition('melt', solid, liquid)
machine.melt()

Traceback (most recent call last):
  File "<input>", line 5, in <module>
  File "/home/stefanga/.virtualenvs/reactor/lib/python3.4/site-packages/transitions/core.py", line 531, in add_transition
    for s in source:
TypeError: 'State' object is not iterable
```

A possible workaround right now is to create "constants" for the state names and use them, but the example above feels much more object oriented and I was suprised that it didn't work.

tyarkoni:
Addresses issue in #153--allows `State` instances to be passed in as arguments to `add_transition` instead of requiring strings.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -521,6 +521,7 @@ class Machine(object):
 
         if isinstance(source, string_types):
             source = list(self.states.keys()) if source == '*' else [source]
+        source = [s.name if isinstance(s, State) else s for s in listify(source)]
 
         if self.before_state_change:
             before = listify(before) + listify(self.before_state_change)
@@ -529,6 +530,8 @@ class Machine(object):
             after = listify(after) + listify(self.after_state_change)
 
         for s in source:
+            if isinstance(dest, State):
+                dest = dest.name
             t = self._create_transition(s, dest, conditions, unless, before,
                                         after, prepare, **kwargs)
             self.events[trigger].add_transition(t)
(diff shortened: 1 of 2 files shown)


transitions/core.py
```diff
@@ -521,6 +521,7 @@ def add_transition(self, trigger, source, dest, conditions=None,
 
         if isinstance(source, string_types):
             source = list(self.states.keys()) if source == '*' else [source]
+        source = [s.name if isinstance(s, State) else s for s in listify(source)]
```
aleneum: Nice one! You could put that in an `else` statement to prevent double `listify`

transitions/core.py
```diff
@@ -529,6 +530,8 @@ def add_transition(self, trigger, source, dest, conditions=None,
             after = listify(after) + listify(self.after_state_change)
 
         for s in source:
+            if isinstance(dest, State):
+                dest = dest.name
```
tyarkoni: Wait, no--that's checking `dest`, not `source`. At this point `dest` hasn't been converted, only `source`.

aleneum: Argh, my bad. you are right

transitions/core.py
tyarkoni: Done

PR #158, opened 2016-11-14
Allow injecting lock instance (or any arbitrary context manager) into LockedMachine

paulbovbel:
Thanks for the great library!

Small addition, to allow specifying a lock instance, or an arbitrary set of managers in LockedMachine.

### DIFF
--- a/tests/test_threading.py
+++ b/tests/test_threading.py
@@ -107,6 +107,43 @@ class TestLockedTransitions(TestCore):
         self.assertAlmostEqual(blocked - begin, 1, delta=0.1)
 
 
+class TestMultipleContexts(TestCore):
+
+    class TestContext(object):
+        def __init__(self, event_list):
+            self._event_list = event_list
+
+        def __enter__(self):
+            self._event_list.append((self, "enter"))
+
+        def __exit__(self, type, value, traceback):
+            self._event_list.append((self, "exit"))
+
+    def setUp(self):
+        self.event_list = []
+
+        self.c1 = self.TestContext(event_list=self.event_list)
+        self.c2 = self.TestContext(event_list=self.event_list)
+
+        self.stuff = Stuff(machine_cls=MachineFactory.get_predefined(locked=True), extra_kwargs={
+            'context': [self.c1, self.c2]
+        })
+        del self.event_list[:]
+
+        self.stuff.machine.add_transition('forward', 'A', 'B')
+
+    def tearDown(self):
+        pass
+
+    def test_ordering(self):
+        self.stuff.forward()
+        # There are a lot of internal enter/exits, but the key is that the outermost are in the expected order
+        self.assertEqual(self.event_list[0], (self.c1, "enter"))
+        self.assertEqual(self.event_list[1], (self.c2, "enter"))
+        self.assertEqual(self.event_list[-2], (self.c2, "exit"))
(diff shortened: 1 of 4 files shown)


aleneum:
Hi @paulbovbel,

sounds like a good idea and thank you for the pull request. I will review it as soon as I can.

Best regards,
Alex

transitions/extensions/locking.py
```diff
@@ -3,36 +3,64 @@
...
+        except KeyError:
+            pass
+
+        try:
+            self.managers.extend(kwargs.pop('managers'))
+        except KeyError:
+            pass
+
+        if not self.managers:
+            self.managers.append(threading.RLock())
```
aleneum: `RLock` is imported from `threading`. `self.managers.append(RLock())` should be sufficient.

transitions/extensions/locking.py
```diff
@@ -3,36 +3,64 @@
...
+        try:
+            self.managers.append(kwargs.pop('lock'))
+        except KeyError:
+            pass
+
+        try:
+            self.managers.extend(kwargs.pop('managers'))
+        except KeyError:
+            pass
+
```
aleneum: Having two keywords which are pretty much handled the same way might be a bit confusing. I do not know which word might encompass both (`context` maybe?).

transitions/extensions/locking.py
```diff
@@ -3,36 +3,64 @@
...
 
     def __init__(self, *args, **kwargs):
-        self.rlock = RLock()
+        self.managers = []
+
+        try:
+            self.managers.append(kwargs.pop('lock'))
+        except KeyError:
+            pass
+
```
aleneum: Since you already use contextlib, we could also make use of `suppress` or [Jeff Paine's
 fallback](https://gist.github.com/JeffPaine/6213790#better-23) to shorten this a bit. In the longer run both `nested` and `suppress` might go to a `utils` module (together with `listify`) to clean up the code. If we end up with just one keyword (instead of `lock` and `managers`) this might be a bit over the top though.

transitions/extensions/locking.py
paulbovbel: agreed

paulbovbel:
Interface fixed, tests pending, thanks for review

paulbovbel:
test added

tests/test_threading.py
```diff
@@ -107,6 +107,43 @@ def test_pickle(self):
...
+        del self.event_list[:]
+
+        self.stuff.machine.add_transition('forward', 'A', 'B')
+
+    def tearDown(self):
+        pass
+
+    def test_ordering(self):
+        self.stuff.forward()
+        # There are a lot of internal enter/exits, but the key is that the outermost are in the expected order
```
aleneum: Yes and no. Yes: Locks have to be reentrant since every function which can be called will require the look and this procedure is most likely nested. No: Afaik there is no monkey patching going on, only inheritance and therefore method overloading/overriding. What could be done is to check wether the lock has been acquired already and if that is the case skip the `LockedMethod` wrapping and return the base class method directly. This needs to be done for `LockedEvent` as well. I will have a look into it.

aleneum:
The limitation to reentrant locks is no show stopper imho. Thank you for suggesting and implementing this features @paulbovbel. I created an issue for the redesign of the lock mechanism to also support other types of locks (#167).

Issue #165, opened 2016-12-07
Single state machine, multiple models


paulbovbel:
I may be missing something, but it's definitely possible to:
- create a state machine with an internal model
- create a state machine that managers one or more external models?

Is there any way to define a state machine once, and register model(s) to it after the fact?

Any guidance on the feasibility of implementing something like that?

tyarkoni:
This can't be done right now (well, you *could* just_ append to `machine.models`, and that will give you everything except the ability to register triggers properly), but it makes sense to move the model creation code into an encapsulated `Machine.add_model()` method, and it would be very straightforward.

tyarkoni:
Closed in #166.

PR #166, opened 2016-12-07
Separate add_models method for core.Machine

paulbovbel:
In the interest of having a single(ton) state machine definition apply repeatedly to temporary model instances, I've tried to adapt Machine to allow adding models after the fact. (see #165)

This works fine for the core tests, but appears to break anything nested related, so more work from me is required. If anyone with experience in the 'nested' extension can give some insight, it would be appreciated!

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -298,7 +298,7 @@ class Machine(object):
                  send_event=False, auto_transitions=True,
                  ordered_transitions=False, ignore_invalid_triggers=None,
                  before_state_change=None, after_state_change=None, name=None,
-                 queued=False, **kwargs):
+                 queued=False, add_self=True, **kwargs):
         """
         Args:
             model (object): The object(s) whose states we want to manage. If None,
@@ -337,6 +337,7 @@ class Machine(object):
                 executed in a state callback function will be queued and executed later.
                 Due to the nature of the queued processing, all transitions will
                 _always_ return True since conditional checks cannot be conducted at queueing time.
+            add_self (boolean): If no model(s) provided, intialize state machine against self.
 
             **kwargs additional arguments passed to next class in MRO. This can be ignored in most cases.
         """
@@ -380,11 +381,11 @@ class Machine(object):
             self.add_ordered_transitions()
 
         if model:
-            self.add_models(model)
-        else:
-            self.add_models(self)
+            self.add_model(model)
+        elif add_self:
+            self.add_model(self)
 
-    def add_models(self, model):
+    def add_model(self, model):
         models = listify(model)
 
         for model in models:
@@ -400,7 +401,7 @@ class Machine(object):
                     self._add_trigger_to_model(trigger, model)
 
                 for _, state in self.states.items():
-                    self._add_state_to_model(state, model)
+                    self._add_model_to_state(state, model)
 
                 self.set_state(self._initial, model=model)
                 self.models.add(model)
@@ -409,7 +410,6 @@ class Machine(object):
         models = listify(model)
 
         for model in models:
-            # TODO remove any callbacks from State to allow models to be GCed
             self.models.remove(model)
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


tyarkoni:
Thanks for the PR! This looks good from my perspective (will take a closer look once tests pass), but we definitely need to make sure the nesting doesn't break. Nesting is @aleneum's baby, so maybe he can weigh in when he has a chance.

One quick question: is there any reason to turn `Machine.models` into a set? I get that this will in principle make `in` checks more efficient, but the list overhead is likely to be trivial here, and I think it's a good idea to maintain a record of the order of entry of the models.

aleneum:
@paulbovbel,

you can use [this](https://github.com/tyarkoni/transitions/files/639036/fix_nested_issues.txt) patch to fix the issues with the nested transitions and graph retrieval. If `patch -p1` does not work I can also fork your fork and create a pull request.

paulbovbel:
Thanks for your patch @aleneum, that worked fantastic.

@tyarkoni I'm torn on the `set()`:

Cons:
- models must be hashable
- no order of entry (is this a use case?)

Pros:
- can't register a model twice with a Machine - should this throw an exception, or is it currently considered valid?
- when removing (TBD) a model from a Machine, there's a clean way to unregister callbacks because the model is guaranteed unique
- efficiency in add/remove

transitions/core.py
```diff
@@ -497,24 +497,30 @@ def add_states(self, states, on_enter=None, on_exit=None,
...
-        if hasattr(model, enter_callback) and \
-                inspect.ismethod(getattr(model, enter_callback)):
-            state.add_callback('enter', enter_callback)
-        exit_callback = 'on_exit_' + state.name
-        if hasattr(model, exit_callback) and \
-                inspect.ismethod(getattr(model, exit_callback)):
-            state.add_callback('exit', exit_callback)
+        for trigger in ('enter', 'exit'):
+            try:
+                state.add_callback(trigger, self._lookup_callback(model, state, trigger))
```
paulbovbel: I had initially intended to implement a state.remove_callback, until I realized that this is adding a string and not an actual function callback (which would have created a reference) to the string.

Is there any harm in leaving stale callback registrations on the state object?

paulbovbel: I'm not sure I understand the purpose of this function, as every time a model is added, if it has the method `model.on_enter_<state>`, state.add_callback will add an instance of the string `"on_enter_<state>"` to the list `State.on_enter`

transitions/core.py
```diff
@@ -334,6 +337,7 @@ def __init__(self, model=None, states=None, initial=None, transitions=None,
                 executed in a state callback function will be queued and executed later.
                 Due to the nature of the queued processing, all transitions will
                 _always_ return True since conditional checks cannot be conducted at queueing time.
+            add_self (boolean): If no model(s) provided, intialize state machine against self.
```
tyarkoni: Not sure I like the idea of adding an `add_self` argument. I agree that we don't want to assume the user wants to add `self` to the model list automatically, but I think I'd prefer to handle this by making the default value of `None` leave the model list empty, and requiring a subsequent `add_model()` call. We could have a special case for the string value `self`--i.e., `Machine(model='self')` would produce the current behavior.

This isn't ideal, obviously, because it would break the API. But once we allow dynamic addition of models, it starts to seem quite counterintuitive to me to that the way to obtain a truly empty model list is to explicitly pass `add_self=False`. It seems like the more sensible approach is to default to an empty list, and make the user explicitly indicate if they want to populate the model list with the Machine instance itself.

Maybe we can keep the proposed approach for now, and save the API-breaking change for a major version release. Thoughts?

paulbovbel:
Looks like the set() implementation wasn't necessary, so happy to unwind.

transitions/core.py
```diff
@@ -376,12 +380,37 @@ def __init__(self, model=None, states=None, initial=None, transitions=None,
...
-                logger.warning("%sModel already contains an attribute 'trigger'. Skip method binding ",
-                               self.id)
-            else:
-                model.trigger = partial(get_trigger, model)
+        if model:
+            self.add_model(model)
+        elif add_self:
+            self.add_model(self)
+
+    def add_model(self, model):
```
tyarkoni: Please add a docstring.

tyarkoni: Actually, it looks like some of the other public methods also don't have docstrings, so you can ignore this and I'll do a pass later.

tyarkoni:
@paulbovbel I don't have strong feelings about the set thing, but since it's not necessary, and performance is not likely to be an issue, I think we should probably stick with a list. I can't think of a use case for model order info off the top of my head, but it doesn't hurt to have it, seeing as the logic is otherwise identical.

transitions/core.py
paulbovbel: +1, API compatibility is king

paulbovbel:
set reverted to list

transitions/core.py
paulbovbel: added

transitions/core.py
wtgee: Since a `model=None` currently adds `self` to the list I don't understand why you wouldn't also just want that functionality here.

tyarkoni: @wtgee this PR adds the ability to dynamically add models after `Machine` initialization. Before, it made no sense to initialize with an empty model list, so `None` could be interpreted as "use yourself as model". Now, the user may want to initialize with no models, and add some later, so if we wanted to avoid using an `add_self` argument, we'd need to change the behavior of `model=None`. But I agree that for the time being, we should avoid breaking the API. So I'll merge the current version as-is.

tyarkoni:
Looks good, thanks again!

transitions/core.py
wtgee: @tyarkoni Makes sense. For the future I still think we should assume most uses will have a single model and opt not to break the API, but we can figure that out at that point and don't need to go too much into it here.

Thanks @paulbovbel! Nice clean implementation.

PR #171, opened 2016-12-14
Use function name when printing the condition

cemoody:


aleneum:
good call. thanks @cemoody

Issue #167, opened 2016-12-08
Extend lock injection to support simple locks


aleneum:
Currently, only reentrant locks are supported by design. A solution would be to check if a lock has been already acquired by the current thread and skip the `LockedMethod/Event` wrapping in this case. Even though `checking` is probably what a RLock is doing anyway this might improve the performance due to the nested nature of method calls. Additionally `transitions` becomes for versatile concerning its interaction with other multithreaded components.

paulbovbel:
In my use case, I'm using re-entrant context managers (an instrumented RLock and django.transaction.atomic), but because of the event frequencies involved, the overhead of triggering the re-entrant behaviour (django generating a save point) ~10+ times for a single trigger is unpleasant.

aleneum:
In `dev-0.5`, we have introduced a custom context manager which will keep track of the thread that currently entered a `with` block. Entering a second time will be omitted. The default locking mechanism still is a `threading.RLock` since `threading.Lock` seems to have issues with pickling.
Additionally, `LockedMethod` wrapping has been reduced to 'public' methods (every method not starting with an underscore) to further reduce the locking overhead.
Considering this simple example

```python
m = M(states=['A', 'B', 'C', 'D'], transitions=[['reset', '*', 'A']], initial='A')
m.get_triggers('A')
```

the following has been achieved:

* reduced execution time from ~600% to ~400% (in contrast to `transitions.Machine`)
* reduced maximum stack size of `with` blocks from 3 to 1
* reduced entering `with` blocks  attempts from 72 (!) to 4

paulbovbel:
Thanks! Looking forward to 0.5

PR #172, opened 2016-12-15
Add per-model handling of initial states and locks

paulbovbel:
- Allow not specifying initial state in machine.__init__ when no models are being added
- Allow specifying initial state in machine.add_model, optional when initial specified in init
- Allow specifying per-machine and per-model contexts in LockedMachine. Current default behaviour of one per-machine global RLock() maintained.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -359,15 +359,16 @@ class Machine(object):
         self._transition_queue = deque()
         self.models = []
 
-        if initial is None:
-            self.add_states('initial')
+        if model is None and add_self:  # TODO(pbovbel) reconsider API for next major release
+            model = self
+
+        if model and initial is None:
             initial = 'initial'
+            self.add_states(initial)
         self._initial = initial
 
         if states is not None:
             self.add_states(states)
-
-        self.set_state(self._initial)
 
         if transitions is not None:
             transitions = listify(transitions)
@@ -382,12 +383,16 @@ class Machine(object):
 
         if model:
             self.add_model(model)
-        elif add_self:
-            self.add_model(self)
 
-    def add_model(self, model):
+    def add_model(self, model, initial=None):
         """ Register a model with the state machine, initializing triggers and callbacks. """
         models = listify(model)
+
+        if initial is None:
+            if self._initial is None:
+                raise MachineError("No initial state configured for machine, must specify when adding model.")
+            else:
+                initial = self._initial
 
         for model in models:
             if model not in self.models:
@@ -404,7 +409,7 @@ class Machine(object):
                 for _, state in self.states.items():
                     self._add_model_to_state(state, model)
 
-                self.set_state(self._initial, model=model)
+                self.set_state(initial, model=model)
                 self.models.append(model)
 
     def remove_model(self, model):
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -52,9 +52,6 @@ class State(object):
         self.on_enter = listify(on_enter) if on_enter else []
         self.on_exit = listify(on_exit) if on_exit else []
 
-    def __str__(self):
-        return self.name
-
     def enter(self, event_data):
         """ Triggered when a state is entered. """
         logger.debug("%sEntering state %s. Processing callbacks...", event_data.machine.id, self.name)
--- a/README.md
+++ b/README.md
@@ -729,6 +729,40 @@ lump.state
 
 Here you get to consolidate all state machine functionality into your existing model, which often feels more natural way than sticking all of the functionality we want in a separate standalone `Machine` instance.
 
+You can also create a standalone machine, and register models dynamically via `machine.add_model`. Remember to call `machine.remove_model` if machine is long-lasting while your models are temporary and should be garbage collected:
+
+```python
+class Matter():
+    pass
+
+lump1 = Matter()
+lump2 = Matter()
+
+machine = Machine(states=states, transitions=transitions, initial='solid', add_self=False)
+
+machine.add_model(lump1)
+machine.add_model(lump2, initial='liquid')
+
+lump1.state
+>>> 'solid'
+lump2.state
+>>> 'liquid'
+
+machine.remove_model([lump1, lump2])
+del lump1  # lump1 is garbage collected
+del lump2  # lump2 is garbage collected
+```
+
+If you don't provide an initial state in the state machine constructor, you must provide one every time you add a model:
+
+```python
+machine = Machine(states=states, transitions=transitions, add_self=False)
+
+machine.add_model(Matter())
+>>> "MachineError: No initial state configured for machine, must specify when adding model."
+machine.add_model(Matter(), initial='liquid')
+```
+
 ### Logging
 
 Transitions includes very rudimentary logging capabilities. A number of events--namely, state changes, transition triggers, and conditional checks--are logged as INFO-level events using the standard Python `logging` module. This means you can easily configure logging to standard output in a script:
@@ -988,7 +1022,7 @@ thread.start()
 machine.new_attrib = 42 # not synchronized! will mess with execution order
 ```
 
-Any python context manager can be passed in via the `context` keyword argument:
+Any python context manager can be passed in via the `machine_context` keyword argument:
 
 ```python
 from transitions.extensions import LockedMachine as Machine
@@ -999,7 +1033,16 @@ states = ['A', 'B', 'C']
 lock1 = RLock()
 lock2 = RLock()
 
-machine = Machine(states=states, initial='A', context=[lock1, lock2])
+machine = Machine(states=states, initial='A', machine_context=[lock1, lock2])
+```
+
+Any contexts via `machine_model` will be shared between all models registered with the `Machine`.
+Per-model contexts can be added as well:
+
+```
+lock3 = RLock()
+
+machine.add_model(model, model_context=lock3)
 ```
 
 It's important that any user-provided context managers are re-entrant since the state machine will call them multiple
--- a/tests/test_add_remove.py
+++ b/tests/test_add_remove.py
@@ -3,7 +3,7 @@ try:
 except ImportError:
     pass
 
-from transitions import Machine
+from transitions import Machine, MachineError
 from unittest import TestCase
 
 import gc
@@ -51,3 +51,51 @@ class TestTransitionsAddRemove(TestCase):
         s3.advance()
         s3.advance()
         self.assertTrue(s3.is_C())
+
+    def test_add_model_initial_state(self):
+        states = ['A', 'B', 'C', 'D', 'E', 'F']
+        machine = Machine(states=states, initial='A', name='Test Machine', add_self=False)
+        machine.add_transition('advance', 'A', 'B')
+        machine.add_transition('advance', 'B', 'C')
+        machine.add_transition('advance', 'C', 'D')
+
+        s1 = Dummy()
+        s2 = Dummy()
+        s3 = Dummy()
+
+        machine.add_model(s1)
+        machine.add_model(s2, initial='B')
+        machine.add_model(s3, initial='C')
+
+        s1.advance()
+        s2.advance()
+        s3.advance()
+
+        self.assertTrue(s1.is_B())
+        self.assertTrue(s2.is_C())
+        self.assertTrue(s3.is_D())
+
+    def test_add_model_no_initial_state(self):
+        states = ['A', 'B', 'C', 'D', 'E', 'F']
+        machine = Machine(states=states, name='Test Machine', add_self=False)
+        machine.add_transition('advance', 'A', 'B')
+        machine.add_transition('advance', 'B', 'C')
+        machine.add_transition('advance', 'C', 'D')
+
+        s1 = Dummy()
+        s2 = Dummy()
+        s3 = Dummy()
+
+        with self.assertRaises(MachineError):
+            machine.add_model(s1)
+        machine.add_model(s1, initial='A')
+        machine.add_model(s2, initial='B')
+        machine.add_model(s3, initial='C')
+
+        s1.advance()
+        s2.advance()
+        s3.advance()
+
+        self.assertTrue(s1.is_B())
+        self.assertTrue(s2.is_C())
+        self.assertTrue(s3.is_D())
--- a/tests/test_threading.py
+++ b/tests/test_threading.py
@@ -141,7 +141,7 @@ class TestMultipleContexts(TestCore):
         self.stuff.machine.add_transition('forward', 'A', 'B')
 
     def tearDown(self):
-        pass
+        self.stuff.machine.remove_model(self.s1)
 
     def test_ordering(self):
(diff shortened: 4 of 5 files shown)


transitions/core.py
```diff
@@ -359,16 +356,17 @@ def __init__(self, model=None, states=None, initial=None, transitions=None,
         self._transition_queue = deque()
         self.models = []
 
-        if initial is None:
-            self.add_states('initial')
+        if model is None and add_self:  # TODO(pbovbel) reconsider API for next major release
+            model = self
+
```
aleneum: I'd suggest to remove `add_self` again. Having accidentally set `add_self=False` and no model leads to weird effects which may be hard to track down. If no model is passed, use `self` as model. If someone wants to explicitly add self AND a model, why not do it in two steps `m = Machine(model); m.add_model(m)` . This keeps the consistency if previous versions and requires one line of code for a special case instead of requiring another parameter to consider for every other case. What do you think?

aleneum:
great addition btw. 👍

transitions/core.py
paulbovbel: Actually, add_self=False and no model is exactly the use case i'm interested in: definining a singleton, long-lived machine object, and then add/remove models (in-memory cache from db data) on the fly.

I also like the possibility of not instantiating a magic `'initial'` state, which is only possible if no models are specified and no `initial` arg is provided.

What kind of weird effects do your foresee?

aleneum: > Actually, add_self=False and no model is exactly the use case i'm interested in

oh, I see. Well, if that is actually a use case I agree that there should be a way to achieve this.

> What kind of weird effects do your foresee?

I was thinking about the easier settings where someone tries to use `Machine` as a model but nothing happens. I am also not a big fan of attributes that depend on other attributes ('adds self but only if...') or change meaning based on the value of other attributes.

Maybe in `transitions` 0.5.x a Machine without model parameter should actually be initialized without a model. Or `add_self` should always add the machine as a model. Or we find a better way to actually pass `Machine` as a part of the model parameter which would be preferable.

> I also like the possibility of not instantiating a magic 'initial' state, **which is only possible** if no models are specified **and no initial arg is provided**.

The state 'initial' is only created if you do not pass `initial` as an argument. You could just pass `initial` to have a default initial argument for all the following models. In your scenario, you would have to pass `initial` to `add_model` anyway to not raise Exceptions. If you plan to initialize `Machine` without states this might be a problem though.

wtgee: While we definitely want this possible I don't think it is the most typical use case. What about just reversing the logic of `add_self` so that there is a `dont_add_self` (with a much better name than that) that would prevent the `Machine` being added as a model but the default is to add `self` so the API doesn't break and people can use a `Machine` in a straight-forward way.

paulbovbel: Agreed re 0.5.x, in the meantime, can we keep add_self to maintain current behaviour/API but enable my use case?

Sometimes it doesn't make sense to provide an 'initial' state. In my case, I'm loading/unloading models all the time as they get cached from the db, meaning their 'initial' state is actually whatever the DB says.

I think the MachineError here makes sense as currently implemented (if no default 'initial' and no 'initial' in add_model) but i'm totally open to alternatives.

tyarkoni:
I agree, this is an excellent addition, thanks! Re: the `add_self` API issue, per discussion in #166, my preference would be to break the API in 0.5 and (a) make it so that `model=None` initializes with an empty model list, and (b) remove the `add_self` argument entirely. Then users who want to initialize with `self` as the model can pass the special string 'self'--i.e., `Machine(model='self')`. I think that approach has the most sensible semantics, as I agree with @aleneum that it's kind of weird to say "hey, set yourself as the model by default, unless you explicitly say that you don't want that to happen". It makes much more sense to me to make the user explicitly acknowledge that they're trying to do something a bit counterintuitive (though very useful), which is to set the newly initialized `Machine` as its own model. That also means a naive user who hasn't bothered to read the documentation can probably intelligently guess (correctly) that the following will successfully set the `Machine` as its own model:

```python
m = Machine(...)
m.add_model(m)
```

transitions/core.py
aleneum: > Agreed re 0.5.x, in the meantime, can we keep add_self to maintain current behaviour/API but enable my use case?

sure, why not. with your additions, 0.5.0 is around the corner anyway :).

> I think the MachineError here makes sense as currently implemented (if no default 'initial' and no 'initial' in add_model)

yes, it does make sense if we allow Machine to not have a 'default' initial state. I am not sold on the idea we should let that happen.
But maybe we find a way to have `None` as a valid value for the initial parameter while keeping the current behaviour as default (e.g. `initial=False`?). If someone tries to intialize a Machine with a model and with `initial=None`an Exception is raised.

aleneum:
@tyarkoni:

yeah, I was also thinking about 'self'. What do you (@all) prefer as the default value then? model=None or model='self'? I can't decide. Second would stick to the current behaviour and also prominently show the availability of this option. First would be cleaner I guess.

transitions/core.py
paulbovbel: That might be getting too clever:

```
initial=False - no initial state, Error if models provided or add_self=True in __init__

initial=None - default, create 'initial' state for machine

initial=something - must be a valid state in machine
```

Torn whether implementing that is worthwhile rather than just enforcing that SOME initial state is provided. However, that would allow initialising a machine with **_nothing_**, and then dynamically adding states, transitions and models piecemeal.

aleneum: The documentation never mentions that an 'initial' state is created if initial=None if I am not mistaken. Maybe we should just remove that behaviour completely. What do you think @tyarkoni and @wtgee?

tyarkoni:
Re: 'self'--funny, it didn't occur to me that we could just make `model='self'` the default, in which case the current behavior is preserved. I think by far the most common use case is that `model` will be explicitly passed another object or list anyway, so it's not like making `model='self'` the default will add work for most people (only the few who really want to initialize with no model at all). So I lean towards 'self' being the default. And actually, if we go with this solution, I would move to change this ASAP in the new release so that we can drop `add_self` before many people get used to it.

Re: `initial`, I guess I have a pretty strong aversion to forcing users to specify the `initial` argument in the most common use case where the `Machine` doesn't really start doing anything useful until it's explicitly triggered the first time. I think part of the reason the package has succeeded is that it appears, to the eye, very minimalistic (though there's plenty of power under the hood), and its behavior can be guessed at without having to read very much documentation. While conceptually it's perfectly sensible to expect a state machine to require a user to specify what state it starts in, I like packages that make sane guesses about the defaults. I also suspect that removing the automatic introduction of an 'initial' state when needed will break a *lot* of people's code. While I'm not averse to that on principle, this one strikes me as a change that we can easily avoid.

That said, I definitely agree with @paulbovbel that there should be some ability to initialize without a default `initial` state. I agree that the suggestion to allow `None`, `False`, or `str` values is a bit clunky, but I think it's probably the best balance between supporting the functionality we want and maintaining users' goodwill. I would probably reverse the meaning of `None` and `False` though, so that the new default is `False`, and it behaves like the current `None`--i.e., it creates a new 'initial' state if none is provided. Then explicitly setting `None` would produce the behavior @paulbovbel wants, where an exception is thrown if models are provided. One reason I like that is that it would be consistent with how we're handling the `model` argument, in that we're cultivating the expectation that some of the key arguments sometimes have to be explicitly set to `None`.

What do you guys think?

paulbovbel:
defaults that aren't `None` can have some caveats, so I would propose two options:

Option 1.

`initial` (default `None`):
if None, make `'initial'` state under the hood. if `False`, set no initial state. if `'something'`, make `'something'` initial state, must be added in `states`.

`model` (default `None`):
if `None`, add `self`. Otherwise, listify and add as model. If `[]`, would add no models.

-----

Option 2.

`initial` (default `'initial'`):
if `'initial'`, make `'initial'` state under the hood. if falsey (`None` or `False`), set no initial state. if `'something'`, make `'something'` initial state, must be added in `states`.

`model` (default `'self'`):
if `'self'`, add `self`. Otherwise, listify and add as model. If falsey (`None` or `[]), would add no models.

----

Let me know which option is preferable, and I will implement as part of this PR.

aleneum:
> if 'initial', make 'initial' state under the hood. 

hmm... what about default='initial' and 'initial' and 'something' get the same treatment:
If 'something' not in self.states, create state 'something'. This adds a possible bug based on spelling though but removes an execution side track. Otherwise we have to check if 'initial' may already exist.

paulbovbel:
I guess I see an arg that takes a 'placeholder' like a string and then does something else as dynamic, but that can be argued either way.

Anyways, will update PR with Option 2-ish.

tyarkoni:
+1 to @aleneum's suggestions. And actually, we should probably already be checking if a `State` with name 'initial' exists in the `states` argument before calling `Machine.add_states('initial'), because it's conceivable that a user will want to define that state themselves and give it some extra behavior. I would suggest maybe deferring the addition of the 'initial' state until *after* the other states have been added. That will allow us to check against the keys in the `.states` dict, and only add the `initial` state if it doesn't exist.

This also brings up a more general issue: right now, no uniqueness checking is done in `Machine.add_states`, which has the potentially undesirable side effect of overwriting older states with newer ones that share the same name. I think it makes sense to enforce a name uniqueness constraint, so maybe we should handle this by raising an exception if the user is trying to add a `State` whose name already exists. We could add an explicit `overwrite` argument that would suppress the exception and give the current behavior (where the last `State` passed in with a given name is used).

paulbovbel:
Not updating this PR.

I see a lot of tests and README docs that support:

```
states = ['beginning', 'middle', 'end']
m = Machine(None, states...)
```

Which breaks if meaning of model=None changes.

So please evaluate this PR on the merits of not changing the API (where add_self has already gone in), and leave API-refactoring to another contribution.

aleneum:
@paulbovbel:

> I see a lot of tests and README docs that support... Which breaks if meaning of model=None changes...  leave API-refactoring to another contribution.

fair enough. What about this: You close this PR and open another one to merge your changes into `dev-0.5`? I would also use this branch to tackle #167. I (or someone else) will then do a PR from `dev-0.5` to `master` during the next two weeks where we can evaluate if your uses cases will be covered.

paulbovbel:
Is this current PR unacceptable as-is? It doesn't modify the API ('add_self' already exists, initial=None behaviour is maintained).

I know an API shakedown tends to drag on, and I'm really keep on adding an 'initial state' arg to 'add_model' so I can roll transitions into my project.

aleneum:
> Is this current PR unacceptable as-is?

Not at all! I (or we) meant it when I/we said thats a really neat contribution!

I can only speak for myself when I say that the only reason why I hesitate is that I fear that you have to code stuff twice if we change the meaning of keywords later on. The features you request are valid and reasonable. That's why I would like to see them merged into the `master` asap. But I do not want to change them much when they have been added there because it will require users to alter their code.

Anyway, if `dev-0.5` is no option for you, how about this(@all):
we add these features (as is) to 0.4.4 since they only add new configuration scenarios.
Additionally, we add deprecation warnings for model=None and initial=None. This may also be good practise since it warns users about the upcoming changes.

paulbovbel:
+1, if that's acceptable. I have no problem updating my internal glue code later if there's an API shift for 0.5.

aleneum:
Alright, if @tyarkoni agrees, I will merge your PR and later add deprecation warnings for the 0.4.4 release. I was considering a setting like this (sorry for the mess. I am a bit in a hurry.):

```python
class Machine:

    def __init__(self, *args, **kwargs):
        # ToDo: mapping of args to kwargs
        if model in kwargs and model is None:
            warning.warn("Beginning with release 0.5 model=None will not add any model to the machine. Use model='self' instead.",
                         DeprecationWarning)
        if add_self in kwargs:
           warning.warn("Beginning with release 0.5 add_self will be removed in favour for model='self'.",
                          DeprecationWarning)
        if  model is None and add_self:
            kwargs['model'] = 'self' # remove this for release 0.5

        self._future_init(**kwarg)      

     def _future_init(model='self', ...)
         ....
```

similar behaviour for initial. I am not sure if this is how its done in Python but the above mentioned approach (masking the actual calls and checking the passed values) has the advantage that users who have not tinkered with model=None or initial=None, will see no warnings. For all of those people we can switch the default values without any interruption.
The rest will have one release to adapt their code before it will function differently. The warnings should still be there until release 0.5.1 at least. Feedback welcome!

wtgee:
:+1: Looks good.

@aleneum there is a built-in PendingDeprectationWarning that is typically used: https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning

tyarkoni:
I'm on board with the deprecation warning, but I really don't like the idea of implicitly passing `args` and `kwargs` in `init`, because it makes it impossible to know what the valid arguments are without inspecting the code or reading the docstring. As someone who relies on IDE completions a lot to determine what arguments to pass, I'm loathe to force users to go read the docs every time they want to initialize a `Machine` and can't remember what the valid arguments are.

I don't see it as a serious problem that users who might not have explicitly specified `model=None` will still get a warning, because some of those users will no doubt be intending to use the `Machine` as `model`, but simply not explicitly stating that because they know it's the default. So I think we still want to make it clear that the behavior will change even if users haven't tampered with the default arguments.

aleneum:
>  As someone who relies on IDE completions a lot to determine what arguments to pass...

good point. totally forgot about that (event though I make use of this feature a LOT as well). No real option, I agree.

> I don't see it as a serious problem that users who might not have explicitly specified model=None will still get a warning, because some of those users will no doubt be intending to use the Machine as model, but simply not explicitly stating that because they know it's the default.

Agree, but according to my current understanding this wont change in the future since we switch from model=None/add_self=True to model='self' with exactly the same behaviour. So setting the machine as model will still be the default (or won't it?). Its kind of tough to communicate in a warning that you do not have to change anything. It might startle people without a reason.

We could just replace model=None with model='self' as the default value for `0.4.4`. In this case warnings are only shown to people which set model=None explicitly. model=None and add_self=True will of course be treated like before. Is that an option?

First warning suggestion. Will be shown if `model is None and add_self is True`:
"Starting from transition version 0.5.0, passing model=None to the constructor will no longer add the machine instance as a model but add NO model at all. Consequently, add_self will be removed. To add the machine as a model (and also hide this warning) use the new default value model='self' instead."

paulbovbel:
> We could just replace model=None with model='self' as the default value for 0.4.4. In this case warnings are only shown to people which set model=None explicitly. model=None and add_self=True will of course be treated like before. Is that an option?

+1

paulbovbel:
Ping for comment and/or merge and/or release

aleneum:
I am willing to merge it and conduct all the previously mentioned changes, but let's see what @tyarkoni  and/or @wtgee think.

tyarkoni:
Thanks for the ping--and yes, I agree with @aleneum's suggestion too. I think as soon as `model='self'` is the default, this PR is ready to merge.

wtgee:
:+1:

PR #176, opened 2017-02-02
Dev 0.4.4

aleneum:
I added the `PendingDeprecationWarning` for the constructors. However, I stumbled upon two issues:

* catching warnings in tests does not work as expected in python 3.3.x
* enabling warnings also enables `DeprecationWarnings` for `contextlib.nested`

The first issue I ‘solved’ by not executing the test if tested with python 3.3. I admit this is rather hacky but I see the `PendingDeprecationWarning` when I create a `Machine` in a console and cannot find a bug report or something related to this.

The second issue is ‘solved’ for now by suppressing DeprecationWarnings in general if `contextlib.nested` is used (part of the locking extension). 
In [this SO post](http://stackoverflow.com/questions/16083791/alternative-to-contextlib-nested-with-variable-number-of-context-managers) it is discussed to replace nested with a backport of `ExitStack`. I think this is a valid solution but it forces us to either ship an implementation of ExitStack or adding an optional dependency (contextlib2 for Python 2.7) which contains a backport of ExitStack. 

What is your oppinion about these things?


wtgee:
:+1: I just ran this locally against some of my machines and all looks good. I think the issues concerning the warnings are fine in the testing as long as we know they get displayed to the users properly.

aleneum:
Nice, thanks.

I think the ExitStack/contextlib.nested issue is a matter of its own. This should probably discussed when we plan to something about it.

PR #178, opened 2017-02-05
fix(test): skip tests relying on *pygraphviz* if not installed

ankostis:
It simply does not fail when PyGraphiviz is not installed.

This is needed on Windows because there is not yet a *wheel* for python-3.5+

### DIFF
--- a/tests/test_threading.py
+++ b/tests/test_threading.py
@@ -9,6 +9,7 @@ import logging
 
 from transitions.extensions import MachineFactory
 from transitions.extensions.nesting import NestedState
+from unittest import skipIf
 from .test_nesting import TestTransitions as TestsNested
 from .test_core import TestTransitions as TestCore
 from .utils import Stuff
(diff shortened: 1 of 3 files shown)


tests/test_threading.py
```diff
@@ -168,6 +175,7 @@ def test_model_context(self):
 
 
 # Same as TestLockedTransition but with LockedHierarchicalMachine
+@skipIf(pgv is None, 'AGraph diagram requires pygraphviz')
```
aleneum: afaik, no graph functionality is tested here. Instead of `skipIf` you can just remove `graph=True` from `get_predefined` in line 176(184). This way we do not have to pull pygraphviz into test_threading.py.

ankostis: I' not sure how can I locate that line?

A nice point about `unittest.skipIf` is that you get the report of skipped TCs at the end.

aleneum: yeah, I first wondered what 'S' could mean :). The only thing (I assume) you have to do to make the tests in `test_threading` work without pygraphviz is to change this line 
```python
self.stuff = Stuff(states, machine_cls=MachineFactory.get_predefined(locked=True, graph=True, nested=True))
```
into 

```python
self.stuff = Stuff(states, machine_cls=MachineFactory.get_predefined(locked=True, nested=True))
```

in  `setUp` of `TestLockedHierarchicalTransitions`.

PR #180, opened 2017-02-05
feat(core): raise MachineError only for internal-errors

ankostis:
I change certain e xceptiions from `MachineError` either to:
- `ValueError`, e.g. when they are thrown for contructor arg errors), 
- `AttributeErrors`, when invoking a non-existing trigger on the machine

The purpose is for a Machine clients to handle "dynamic" errors that
denote inability to process transition,
differently from "static" setup/configuration errors.

### DIFF
--- a/tests/test_add_remove.py
+++ b/tests/test_add_remove.py
@@ -86,7 +86,7 @@ class TestTransitionsAddRemove(TestCase):
         s2 = Dummy()
         s3 = Dummy()
 
-        with self.assertRaises(MachineError):
+        with self.assertRaises(ValueError):
             machine.add_model(s1)
         machine.add_model(s1, initial='A')
         machine.add_model(s2, initial='B')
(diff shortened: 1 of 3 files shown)


aleneum:
I have no strong oppinion about this. @wtgee, @tyarkoni, what about you?

wtgee:
Could break some current implementation code but other than that it sees like a sensible change.
:+1:

ankostis:
Rebased on top of #178.

ankostis:
Rebased on latest master.

aleneum:
Hi, 

I just want to give you an update why I haven't pulled this request yet. As @wtgee pointed out this might break current implementations catching `MachineError`. I will wait a bit longer to see if @tyarkoni wants to give feedback. I'd assume that the adaptions for the users are bearable and the change clarifies the cause of errors in a reasonable manner. If no one opposes I would merge the changes in a week.

Issue #175, opened 2017-02-02
Why `machine.before` callback happens after `transition.before`?


ankostis:
I think that the order of `machine.before/after` callbacks is a bit asymmetrical.
Currently isolating the order concerning `before/after` events , it looks like that (assuming #174 is correct):

- `transition.before`
- `MACHINE.before`
- `state.exit`
- `STATE CHANGE`
- `state.enter`
- `transition.after`
- `MACHINE.after`

Wouldn't it make sense to "surround" the above block in the machine's default CBs, like that?

- `MACHINE.before`
  - `transition.before`
  - `state.exit`
  - `<STATE CHANGE>`
  - `state.enter`
  - `transition.after`
- `MACHINE.after`

## Rational
I came to question the current ordering because I want to implement a 2-phase commit logic, for all transitions.
So I need to do the 1st phase in `MACHINE.before`, and only finalize it in `MACHINE.after`, if everything was successful.
But any transition-specific login in its `before` cannot access the 1st-phase results (e.g.. a temporary folder, or a branch in some git repo).
And `state.exit` is not specific to the transition (although I can workaround that by checking the destination state from the event-data).

aleneum:
I assume this is just an effect of how it is implemented right now. Imho, it does make sense to call `Machine.before` before `transition.before`. If others agree we can alter the order in no time.

wtgee:
I do think it makes sense to wrap the calls. However I think that might open up the question of what kind of callbacks people will want to write in `MACHINE.before`/`after` such that the question then becomes where do you run the `transitions.prepare`/`conditions`/`unless`. 

Placing the `transition` callbacks inside the `MACHINE`callbacks makes me think that maybe we want the entire `transition` logic called. I'm thinking mostly of something like a custom logger or something else at the `MACHINE` level that would want to help `prepare` a transition or otherwise know about its failure.

ankostis:
(I didn't want to open this can of worms my self... :-)

...but since it's open, I can make a couple more observations:

- Having separate `prepare` methods on each *transition* does not make much sense:
  Why would you "prepare" for a transition that might halt immediately after, in the coming condition?
  A "global" one should suffice.
- I would then couple this new global  `prepare` with a global `finalize` at the end, 
  but only if there is not already some MACHINE at the very end (see case B, below).
- Regarding what you discuss @wtgee , I can see at least 3 possibilities:

Case A: this is my original proposal, where a global `before/after` intermixes with `transition methods:

[...]

Issue #170, opened 2016-12-12
after_state_change functionality


peendebak:
Is there a method to change the `after_state_change` callback after the `Machine` object has been created? And is it possible to have multiple callbacks for `after_state_change`. For example:
```
m = Machine(model=some_model, states=[...])
m.add_callback('after_state_change', callback1)
m.add_callback('after_state_change', callback2)
```

aleneum:
Hello @peendebak,

`after_state_change` can handle lists like this:

```python
machine = Machine(..., after_state_change=['callback1', 'callback2'])
```

Additionally, this list can be adapted during runtime as machine property:

```python
machine.after_state_change = 'callback3'
```

However, this will not have any effect on already created transitions and this is probably what you need right?

peendebak:
@aleneum Thanks for the reply. Your example already helps, but I want the callback to apply to already existing states.

After creating a machine with a state and transitions I want to automatically create a GUI that shows the current state and has options to change the state. If the machine changes state through some other way. I want this to be reflected in the GUI. E.g. after changing the state I want to call `gui.update`. 
I create the gui after creating the machine, so the gui can use `machine.states` to build the interface.

aleneum:
Okay,

to achieve this we could convert `Machine.after/before_state_change` into a property

```python
# Machine
@property
def before_state_change(self):
    return self._before_state_change

# this should make sure that _before_state_change is always a list
@before_state_change.setter
def before_state_change(self, value):
  self._before_state_change = listify(value)
```

skip the concatenation during `add_transition` and iterate over two lists (`Transition.before` and `Machine.before_state_change`) with `itertools.chain`

```python
# Transition.execute
from itertools import chain  
for func in chain(self.before, machine.before_state_change):
    ...
```

Let's see what others think of this or if there might be other solutions.

aleneum:
It took a while but since 6e9334f the following should work:

```python
from transitions import Machine


class Model:

    def bark(self):
        print("bark")

    def chirp(self):
        print("chirp")

model = Model()
m = Machine(model=model, states=['A', 'B'], initial='A', before_state_change='bark')

model.to_B()  # >>> bark
m.before_state_change = 'chirp'
model.to_A()  # >>> chirp
```

PR #184, opened 2017-02-15
Feature #175

aleneum:
* `Machine.before_state_change` and `Machine.after_state_change` will not be appended with `Transition.before/after` any longer.
   - this allows to change `before_state_change` and `after_state_change` during runtime (as required for #170).
* The order of `Machine.before_state_change` and `Transition.before` has been reversed. This way `Machine` methods surround `Transition` methods.
* New constructor parameter `prepare_conditions_check` (temporary name) has been introduced
    - represents a `prepare` for transitions which is called *once* for every triggered event independent of how many transitions might be possible, checked and if these transitions are successful.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -286,14 +286,22 @@ class Event(object):
                 raise MachineError(msg)
         event_data = EventData(state, self, self.machine, model, args=args, kwargs=kwargs)
 
-        for func in self.machine.prepare_conditions_check:
+        for func in self.machine.prepare_transition:
             self.machine._callback(func, event_data)
             logger.debug("Executed machine preparation callback '%s' before conditions." % func)
-        for t in self.transitions[state.name]:
-            event_data.transition = t
-            if t.execute(event_data):
-                return True
-        return False
+
+        result = False
+        try:
+            for t in self.transitions[state.name]:
+                event_data.transition = t
+                if t.execute(event_data):
+                    result = True
+                    break
+        finally:
+            for func in self.machine.finalize_event:
+                self.machine._callback(func, event_data)
+                logger.debug("Executed machine finalize callback '%s'." % func)
+        return result
 
     def __repr__(self):
         return "<%s('%s')@%s>" % (type(self).__name__, self.name, id(self))
@@ -317,9 +325,9 @@ class Machine(object):
 
     def __init__(self, model='self', states=None, initial='initial', transitions=None,
                  send_event=False, auto_transitions=True,
-                 ordered_transitions=False, ignore_invalid_triggers=None, prepare_conditions_check=None,
+                 ordered_transitions=False, ignore_invalid_triggers=None,
                  before_state_change=None, after_state_change=None, name=None,
-                 queued=False, add_self=True, **kwargs):
+                 queued=False, add_self=True, prepare_transition=None, finalize_event=None, **kwargs):
         """
         Args:
             model (object): The object(s) whose states we want to manage. If 'self',
@@ -347,8 +355,6 @@ class Machine(object):
                 that are not valid for the present state (e.g., calling an
                 a_to_b() trigger when the current state is c) will be silently
                 ignored rather than raising an invalid transition exception.
-            prepare_conditions_check: A callable called on for each triggered event before
-                conditions are checked. It receives the very same args as normal callbacks.
             before_state_change: A callable called on every change state before
                 the transition happened. It receives the very same args as normal
                 callbacks.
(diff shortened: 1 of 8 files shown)


aleneum:
Without objection/feedback, I will merge this in a week. This will also be the moment to update the changelog and tag 0.4.4 for a release.

ankostis:
Not an objection, but a wish: 
any thoughts about adding a `finalize_after_all_transitions` in a finally block that fires once, both for successful/failed transitions?

Should I open a new issue to discuss ramifications, such as, where to store collected errors. etc?

aleneum:
> any thoughts about adding a finalize_after_all_transitions in a finally block that fires once, both for successful/failed transitions?

right, I will add this

> Should I open a new issue to discuss ramifications, such as, where to store collected errors. etc?

Please do so. Catching and managing errors is (currently) out of `transitions` scope.

ankostis:
Created #186 for `finalize_after_all_transitions`.

ankostis:
Since you named it `finalize_event`, why not `prepare_event`, to promote the relationship?

aleneum:
> Since you named it finalize_event, why not prepare_event, to promote the relationship?

I thought about this (or something like `on\before_event`) but this might make some people believe it is _always_ executed. But it is only executed if the transition is technically possible (if `state.name in self.transition`). Maybe I am overthinking this though. Phew, naming is tough 😓.

ankostis:
So the prepare/finalize global methods are never executed for "final" states, i.e. those without any outgoing transitions on them.
Ok, that's the expected behavior, I guess.
Or am I missing something here?

Is it possible to have an event, even for a "final" state?

PS: Naming is the hardest part in programming.
But it is fun :-)

aleneum:
> So the prepare/finalize global methods are never executed for "final" states, i.e. those without any outgoing transitions on them.

nope, not right now.

> Is it possible to have an event, even for a "final" state?

well, since we are talking about a feature in development, it's up to us how prepare/finalize behave. We could just extend the try/except block and trigger the prepare and finalize callbacks all the time. In this case I would probably name them `before/after_event` since I feel this is the most consistent scheme considering `before/after_state_change`.

What do you think?

ankostis:
I would expect that for a "final" state, no code should run at all, because there is nothing the user can do to change that.  I mean, it is not that he can try many triggers until the correct one fires - no, there is no correct, trigger, that is a dead-end.  So running code in `prepare/finalize` would be an undesirable side affect, like `opening a db-connection to write empty.

If you care to add such functionality, it might be better to have separate global callback, like `on_dead_trigger` or something like that.   
But that's my 2cnt on this.



Regarding the beauty contest, `before/after` is smooth, but `prepare/finalize`  has the connotation that the last one gets executed regardless of failures.  
And I like the `_event` suffix because it is not related to a single transition, but all those involved (but I guess that's slightly different meaning of the term "event" than yours).

(I hope I'm not intervening too much to your mental space)

aleneum:
> I mean, it is not that he can try many triggers until the correct one fires - no, there is no correct, trigger, that is a dead-end. 

Good point. So I guess we should leave it like it is and only process callbacks if a transition is actually possible.

> prepare/finalize has the connotation that the last one gets executed regardless of failures.

agreed. Also `prepare_transition` does not imply that a transition will actually happen (like `before_transition`) 

> And I like the _event suffix because it is not related to a single transition, but all those involved.

I refer to the underlying `Event` object. Thats why `prepare_event` would not be accurate. But `somewhere_in_event` is no catchy at all. `prepare_transition` because we have already determined that a transition is possible but we do not know which one yet.

ankostis:
By the way, I hadn't  thought of the dead-trigger subtlety.
Now I'm thinking that the proper way to deal with it is to invoke a *single* method on machine(or model), called like `triggered_on_final_state()` that, by default, throws `MachineError`.

Actually I was bitten by this problem while developing my model.  
You see, I'm using YAML-text to setup my transitions, and I had some comma in the wrong place, so an outgoing transitions wasn't declared to the state I wanted (that state was in effect "final"), and I couldn't understand why I was getting `false`, although all conditions were supposed to be `true`.
A `MachineError` would have waked me up.

ankostis:
Setting `ignore_invalid_triggers=False` does the job also for final-states. 
Please ignore my last comment!  It It was late at night and never examine carefully what was my issue back then.

aleneum:
if nobody objects, I will merge this next week. Also this will be the moment to update the manual and the changelog and also tag version 0.4.4.

wtgee:
:+1:

transitions/core.py
```diff
@@ -284,13 +285,27 @@ def _trigger(self, model, *args, **kwargs):
...
+        for func in self.machine.prepare_event:
+            self.machine._callback(func, event_data)
+            logger.debug("Executed machine preparation callback '%s' before conditions." % func)
+
+        result = False
+        try:
+            for t in self.transitions[state.name]:
+                event_data.transition = t
+                if t.execute(event_data):
+                    result = True
```
ankostis: Can we store the `result` on the `event_data.result` so `finalize_event` knows that?
Do we have time before 4.4 release?

PR #195, opened 2017-03-09
doc(proj): add project COORDS on README header and on package

ankostis:
+ Add a table of project "coordinates" as a header on README to quickly view important infos, incl build status.
+ Add more project infos as package attributes (URL, License, Copyright, etc).
+ Added TC to check version/rel-date always in sync with README.md.\
+ Extended travisCI to Python 3.5 & 3.6.


ankostis:
Please check README as rendered in [my branch](https://github.com/ankostis/transitions/tree/projcoords).

aleneum:
It is indeed a good opportunity to add additional information to the ReadMe.
About your recommendations:
* version of current master -- 👍 , a must
* pypi -- also a good call
* meta info in __init__ -- sounds like a good idea
* github -- considering we are talking about the github page... well... not desperately need but okay...
* license -- also a bit redundant but maybe easier to find in the ReadMe.
* update -- hmm.. this sounds like something that may end to be cumbersome to maintain, additionally the information is already on the very same github page; if not steadily maintained 
the two dates may even contradict each other

about the style:

* A table is good but I'd actually prefer a clean line of badges. I will give it a shot with shield.io badges for comparison.

ankostis:
> date -- hmm.. this sounds like something that may end to be cumbersome to maintain

I know.  I have create a TC just for checking this.

Practically, the home and ver/rel-dates infos are needed when downloading archives with the sources locally.  So it really pays the effort of keeping it in syncing with `version.py`.  Besides, how many times you bump version per week?

ankostis:
Forgot to mention that I added a new commit where I extended travisCI to Python 3.5 & 3.6.
All pass OK!

aleneum:
> Forgot to mention that I added a new commit where I extended travisCI to Python 3.5 & 3.6:

right, forgot to mention this. This is also a good addition. We should add the appropriate tags in setup.py as well. 

Checkout the [dev-readme](/tyarkoni/transitions/tree/dev-readme) branch. Imho, this is a bit more polished than a table.

> Besides, how many times you bump version per week?

So this is meant to refer to the last version bump... hmm.. new releases should probably be submitted to PyPi when they are done.  A version bump is not necessarily a new release though. At least that is how I handle it at the moment for transitions. I prefer to bump the version right after a release to show that the current master differs from the release. The 'commits since <release>' badge also adds to this. I do this because in the past, some people were confused because the 0.4.2 from github differed from the PyPi version.

ankostis:
I'm using a different workflow for my projects, where I bump immediately prior to release,
and then bump again to `.dev0` or something like that.
The important thing is that anyone taking the sources outside git, know *roughly* which version he is running.
Our testers need that info, I agree it might not make much sense in this project.

The important thing is to have some idea what are you looking at on the landing page.
Great!

aleneum:
> I'm using a different workflow for my projects 

Usually I do as well. I used to prefer the `master` as a release branch. On github most people open pull requests for the master branch though. Seems like people of github prefer bleeding edge masters.

> The important thing is that anyone taking the sources outside git, know roughly which version he is running

Maybe we should increase the frequency of versioning for minor releases/patches then. Not every version has to end on PyPi though.

> The important thing is to have some idea what are you looking at

👍 

I copy-pasted your changes to `__init__` and `.travis` to the dev branch and also added tags in setup.py and `tox.ini`.

aleneum:
I added additional information about the project into the ReadMe based on your recommendations.
Also, I will strive to encourage more regular minor tags/releases.
Since most addressed points in this pull request have been handled, I close it for now.
Feel free to reopen it, if there is more to discuss.
Thanks again for your input!

PR #197, opened 2017-03-10
Reflexive state notation

janLo:
This adds a new notation to specify reflexive states. It is specially handy if you want to specify reflexive transitions for multiple states at once:
```
machine.add_transition('touch', ['liquid', 'gas', 'plasma'], '=', after='change_shape')
```

This would basically add three transitions from the three states using the `touch` trigger with the source state as destination state.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -659,9 +659,13 @@ class Machine(object):
                 model (e.g., passing trigger='advance' will create a new
                 advance() method in the model that triggers the transition.)
             source(string): The name of the source state--i.e., the state we
-                are transitioning away from.
+                are transitioning away from. This can be a single state, a
+                list of states or an asterisk for all states.
             dest (string): The name of the destination State--i.e., the state
-                we are transitioning into.
+                we are transitioning into. This can be a single state or an
+                equal sign to specify that the transition should be reflexive
+                so that the destination will be the same as the source for
+                every given source.
             conditions (string or list): Condition(s) that must pass in order
                 for the transition to take place. Either a list providing the
                 name of a callable, or a list of callables. For the transition
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -686,9 +686,13 @@ class Machine(object):
             source = [s.name if self._has_state(s) else s for s in listify(source)]
 
         for s in source:
-            if self._has_state(dest):
-                dest = dest.name
-            t = self._create_transition(s, dest, conditions, unless, before,
+            if dest == '=':
+                real_dest = s
+            else:
+                real_dest = dest
+            if self._has_state(real_dest):
+                real_dest = real_dest.name
+            t = self._create_transition(s, real_dest, conditions, unless, before,
                                         after, prepare, **kwargs)
             self.events[trigger].add_transition(t)
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ A lightweight, object-oriented state machine implementation in Python. Compatibl
     - [Transitions](#transitions)
         - [Automatic transitions](#automatic-transitions-for-all-states)
         - [Transitioning from multiple states](#transitioning-from-multiple-states)
+        - [Reflexive transitions from multiple states](#reflexive-from-multiple-states)
         - [Ordered transitions](#ordered-transitions)
         - [Queued transitions](#queued-transitions)
         - [Conditional transitions](#conditional-transitions)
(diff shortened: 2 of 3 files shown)


tyarkoni:
I appreciate the contribution, @janLo, and I can see this being useful on occasion, but I'm not sure it's a sufficiently common use case to warrant adding a special case... the original intent was to keep the core module minimalistic, and I think feature creep is starting to set in. But I could be convinced if others think this is particularly useful. Thoughts?

janLo:
Well, the code is minimal, it's fully tested and documented. The idea came from [fysom](https://pypi.python.org/pypi/fysom). I needed in a project and I thought it's small enough to be done in the library instead of explicitly writing all the states ore having them generated outside.

ankostis:
I agree it is small and useful.  The only drawback is that it prevents from using `=` as a state name (e.g writing a polish machine anyone?).
But in such a fringe case, he would have to overwrite ` add_transition(() method, and still survive. 

I would also give a little thought on the symbol, assuming there are more clever shortcuts in the future, and `=` might be better suited to one of them.  Besides it looks neat.

tyarkoni:
I like the functionality, my concern is about long-term maintainability and feature creep. Each feature like this is individually lightweight and almost always sensible, but can introduce weird edge considerations (cf. @ankostis' comment) and makes the codebase a little harder to learn and maintain. Again, I'm not fundamentally opposed to adding it, but I'd like to hear the other core maintainers' thoughts on it.

tyarkoni:
Also, if we merge it, the behavior should probably be described in the docstring and not just the documentation. (I just noticed that we don't describe what `source='*'` does in the docstring, which was my omission, but that should probably be added as well.)

janLo:
I added both cases to the docstring.

aleneum:
Hi @janLo,

thank you for your contribution. I have seen several occasions where a feature like this may be useful. This being said, I do agree with @tyarkoni that we have to be extra careful when it comes to new features which include either new parameters or symbols.
As mentioned, maintenance  is an issue but also clutter.
We agree on that only if a functionality cannot be achieved by the model or by proper configuration of the machine, a new parameter should be added.
We had this case not long ago when callback processing had to be refined (#175 and related).

'*' representing a wildcard has been around virtually forever but reserving `=` for another mode might interfere with some implementations out there.
I'd suggest `None` as in 'no destination given'. Of course this would still have to be documented but the change to the code would stay very minimalistic.

Its basically just [this](https://github.com/tyarkoni/transitions/blob/master/transitions/core.py#L691) line:

```python
    # create a transition from s to dest, if dest is None, create a reflexive transition
     t = self._create_transition(s, dest or s, conditions, unless, before,
```

janLo:
As said, I choosed `=` because its used in [other projects](https://pypi.python.org/pypi/fysom). that provide state machines for this purpose. I've no preference about the symbol, but I think `None` might cause more confusion than `=` as it has no semantic meaning like "same" or "equal".

In the end its the same as `*` just syntactic sugar. As that I think it should be easy to read and understand.

aleneum:
> In the end its the same as * just syntactic sugar. As that I think it should be easy to read and understand

agree and as already mentioned earlier, the changes are not that big. However it 

1. can be easily done in user code
2. introduces a new parameter/placeholder/attribute
3. and might interfere with current implementations

Having "=" as a state name is valid at the moment while `None` is not. You see I do some rule bending here to work around issue 2 and 3. I am not claiming `None` is the best way to do it, not event that it is a good way. Its a suggestion to reduce new added code but more importantly to not require a reserved character. But yeah, maybe `None` is not as self-explanatory as it should be.

ankostis:
`None` is really not helpful, 
i.e. it could mean both SAME(`'='`) and ALL(`'*'`).  

An alternative would be to use a specific `object` instances:

```python
ALL = object()
SAME = object()
```
instead of strings, but that would eliminate the possibility to configure FSM with text (e.g. I'm using *yaml* in my project).

So I'm in favor of using strings because they are both clear and flexible.

> Having "=" as a state name is valid at the moment while None is not. 

Currently `*` is not a valid name for a source-state, unless it is included in a list, correct?  
In any case, a user can override `Machine.add_transition()` to cancel the special meaning of `'='`.

transitions/core.py
```diff
@@ -686,9 +686,13 @@ def add_transition(self, trigger, source, dest, conditions=None,
             source = [s.name if self._has_state(s) else s for s in listify(source)]
 
         for s in source:
-            if self._has_state(dest):
-                dest = dest.name
-            t = self._create_transition(s, dest, conditions, unless, before,
+            if dest == '=':
+                real_dest = s
```
janLo: This is because if you have more than one source (more than one iteration of the loop there), then after the first loop the value of dest is not = anymore and the destination for all created transitions is the first source state.

janLo:
I fully agree with @ankostis on the preference of strings over objects or None. If conflict with `=` as a possible state name I would go for making it configurable as keyword arguments on initialization of the machine. Like 
```python
def __init__(self, model= ..., wildcard_char='*', same_char='=')
```

transitions/core.py
ankostis: humbled :-)

janLo:
What is the status now? Can this be merged or shall I make the chars configurable?

aleneum:
> I would go for making it configurable as keyword arguments

After some consideration I agree that reflexive transitions fits `transitions` quite well as entering/exiting states are the main hooks to execute code. Especially for model related updates which require certain conditions to be met, this mechanism is useful, generic and adds flexibility which justifies the minor increase in code complexity. We still have to agree on how to handle the 'placeholder' characters though.

Do you think it is required to do that individually for every machine instance?
In `nested` we handle the state separator as a property of the class ([ref](https://github.com/tyarkoni/transitions/blob/master/transitions/extensions/nesting.py#L40)). Same for state and function separator for `Machine` convenience functions ([ref](https://github.com/tyarkoni/transitions/blob/master/transitions/core.py#L342)).
Maybe we could to the same for e.g. `wildcard_all = '*'` and `wildcard_same = '='`.

janLo:
I'm fine with the current state, but I will change it to a class or instance property if preferred. Just tell me what to do to get it merged.

aleneum:
Alright,

if you are okay with it, I will merge it as it is and refactor the chars into variables.

Thank you for your contribution and feedback.

PR #207, opened 2017-03-30
Fixed 'to_' transitions on HSM's

mrdecav:
Have a bug where the 'to_' auto transitions show for nested machines, even when 'show_auto_transitions' is set to false.

Removed equality check for transition events:
... len(event.transitions) == len(self.machine.states)

Seems to make the to_ events get visualised but have no other side effects (I'm sure it was here for a reason, so need some feedback on this change).

### DIFF
--- a/transitions/extensions/diagrams.py
+++ b/transitions/extensions/diagrams.py
@@ -180,8 +180,7 @@ class NestedGraph(Graph):
 
         for event in events.values():
             label = str(event.name)
-            if not self.machine.show_auto_transitions and label.startswith('to_') \
-                    and len(event.transitions) == len(self.machine.states):
+            if not self.machine.show_auto_transitions and label.startswith('to_'):
                 continue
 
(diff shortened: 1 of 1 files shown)


aleneum:
Hi @mrdecav,

the reason for the last check is to distinguish transitions like `to_new_state` from actual auto transitions. Auto transition events usually have as many transitions as states since they should be valid from every state. That's also why the tests fail now. It checks if `m.add_transition('to_state_A', 'B', 'A')` appears in the generated graph.

Could you provide a minimal example where auto transitions are not filtered correctly? I can join the bug hunting then.

aleneum:
Hi @mrdecav,

are you still experiencing the error related to graph generation? If thats so, please provide us with a test case to reproduce the error.

aleneum:
Closing this due to inactivity. If the error still occurs, please let us know and/or reopen this pull request.

PR #210, opened 2017-04-21
update callbacks list in README.md

booware:
changed some callbacks names in 'execution order' section to reflect the real names of the callbacks


aleneum:
Hi @booware,

looks better that way. Thanks.

Issue #133, opened 2016-08-25
GraphMachine cannot plot states with transitions to themselves


chaosentity:
Using: 
Windows 10 64 bit
python 2.7.6 32-bit 
transitions 0.4.1
pygraphviz 1.3.1
graphviz 2.38

I've noticed that if a state machine has a state that can transition to the same state (e.g., state_1 can transition to state_1), pygraphviz encounters an error while trying to plot the state machine using GraphMachine

```python
from transitions.extensions import GraphMachine

states = ['first', 'second', 'last']
transitions = [
    {'trigger': 'move', 'source': 'first', 'dest': 'second'},
    {'trigger': 'go', 'source': 'second', 'dest': 'last'},
    {'trigger': 'end', 'source': 'last', 'dest': 'first'},
    # Uncommenting the following line causes an error in pygraphviz
    # {'trigger': 'nope', 'source': 'first', 'dest': 'first'},
]

class Basic(object): pass

machine = GraphMachine(model=Basic,
                       states=states,
                       transitions=transitions,
                       initial='first',
                       auto_transitions=False,
                       show_conditions=True,
                       )

machine.graph.draw('fsm.png', prog='dot')
```

Uncommenting the 4th transition leads to this error in pygraphviz:
```python
Traceback (most recent call last):
  File "C:/Users/James/PycharmProjects/cmwhitneytest/durrdurrdurr.py", line 24, in <module>
    show_conditions=True,
  File "C:\Python27\lib\site-packages\transitions\extensions\diagrams.py", line 176, in __init__
    super(GraphMachine, self).__init__(*args, **kwargs)
  File "C:\Python27\lib\site-packages\transitions\core.py", line 364, in __init__
    self.add_transition(**t)
  File "C:\Python27\lib\site-packages\transitions\extensions\diagrams.py", line 213, in add_transition
    self.graph = self.get_graph(force_new=True)
  File "C:\Python27\lib\site-packages\transitions\extensions\diagrams.py", line 188, in get_graph
    self.graph = AGraph(self).get_graph(title)
  File "C:\Python27\lib\site-packages\transitions\extensions\diagrams.py", line 154, in get_graph
    self._add_edges(self.machine.events, fsm_graph)
  File "C:\Python27\lib\site-packages\transitions\extensions\diagrams.py", line 122, in _add_edges
    container.add_edge(src.name, dst.name, label=edge_label, ltail=ltail, lhead=lhead)
[...]

PR #217, opened 2017-05-12
Fixed auto_transition discovery in diagrams

KarolOlko:
This isn't much, but this bug actually happened to me, so I've fixed this for posterity :)

Also, at work I use a slightly older version of your library. When I tried to generate a graph on Windows for "\*" transition (that is: source='\*'), I got an error "KeyError: 'agedge: no key'". I could not reproduce it on my Linux PC. Was this a known bug and I should just update the library or you have no knowledge on that and I should investigate further?

### DIFF
--- a/tests/test_graphing.py
+++ b/tests/test_graphing.py
@@ -150,6 +150,13 @@ class TestDiagrams(TestCase):
         self.assertEqual(len(g2.edges()), 4)
         self.assertEqual(len(g2.nodes()), 4)
 
+    def test_trigger_prefixed_with_to_XXX_not_ignored_if_XXX_not_in_states(self):
+        m = self.machine_cls(states=['A', 'B', 'C', 'D'], initial='A', show_auto_transitions=False)
+        m.add_transition('to_end', '*', 'D')
+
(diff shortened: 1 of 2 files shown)


aleneum:
Hello @KarolOlko,

thank you for the pull request. This case hasn't been covered before. 

>  I got an error "KeyError: 'agedge: no key'"

This has been mentioned in #133. Back then it seemed to be an issue with Windows and pygraphviz and could not be reproduced. If you can pinpoint this issue to transitions, feel free to reopen #133 again.

KarolOlko:
Hello @aleneum 

I digged deeper this afternoon. I clone pygraphviz repository on Windows VM and found out that one  examples fails for the same reason as #133  :) This seems to be a graphviz bug. There is precisely the same issue hanging on their bug tracker for more than a year. I won't spend whole weekend trying to compile their sources so I just scraped ~15 lines of code to generate text file input for dot program from Machine class. If I then run dot program from command line, it generates proper output. Good enough for me. If you think this might be a valuable addition for poor Windows souls, I can prepare another PR for that.

PR #218, opened 2017-05-14
Put default value for event data transition `repr` for when transition is not available.

kunalbhagawati:
During the machine's `prepare_event` if we were to inspect the state of the `EventData` instance in a debugger, it will break because the `getattr(self, 'transition')` will return an `AttributeError`.

This fix returns a default value for that.

A default value is return so that the format of the `repr` is conserved and will not lead to confusion.
`[None]` is chosen because the transition name may be any sequence of characters and I felt `None` is a token understood by a lot of people.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -252,7 +252,7 @@ class EventData(object):
 
     def __repr__(self):
         return "<%s('%s', %s)@%s>" % (type(self).__name__, self.state,
-                                      getattr(self, 'transition'), id(self))
+                                      getattr(self, 'transition', '[None]'), id(self))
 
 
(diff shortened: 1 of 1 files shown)


aleneum:
Hi @kunalbhagawati, 

this clearly is a bug. Thank you for pointing it out. I went for initialising `EventData.transition` in the constructor instead of a default getter argument. This way the `transition` argument will be present instantly in the `EventData` object which in my oppinion is a better practise than adding the argument dynamically during transition execution (as it had been before your report). The result is pretty close to your suggestion as `EventData.transition` is initialised with `None`.

PR #220, opened 2017-05-15
Argument matching for callbacks added

KarolOlko:
This was bugging me for a while.
Say I have an trigger and not all callbacks need its arguments (especially conditions, like):
```
class Model(object):
  states = ['unbroken', 'broken']
  def __init__(self, is_concrete=True):
    self.is_concrete = is_concrete
    self.machine = Machine(model=self, (...))
    self.machine.add_transition('hit_hammer', 'unbroken', 'broken',
           conditions='was_hit_hard', unless='is_concrete')
  def was_hit_hard(self, force):
    return force > 50
  def is_concrete(self):
    return self.is_concrete

model.hit_hammer(force=50)
```
. Now the code crashes and I had to resort to changing callback signature to take (\*args, \*\*kwargs), but I would like to avoid that.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -406,6 +406,57 @@ class TestTransitions(TestCase):
         self.assertTrue(m.before_state_change[0].called)
         self.assertTrue(m.after_state_change[0].called)
 
+    def test_machine_matches_event_data_to_callback_signature(self):
+        class TestModel(object):
+            def __init__(self):
+                self.b_mock = MagicMock()
+                self.c_mock = MagicMock()
+
+            def on_enter_B(self):
+                self.b_mock()
+
+            def on_enter_C(self):
+                self.c_mock()
+            
+            def on_enter_D(self, a, b):
+                self.a = a
+                self.b = b
+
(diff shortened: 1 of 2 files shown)


KarolOlko:
Oops.. Now I realized conditions are not covered. This method needs to be moved to EventData. Let me do it in another pull request.

KarolOlko:
Moved function arguments trimming to EventData and checked conditions in unit test added.

tyarkoni:
Can you clarify what the problem behavior is? As detailed in [this README section](https://github.com/tyarkoni/transitions#passing-data), When `send_event=True`, args and kwargs are stored inside the `EventData` object. So the expected pattern here is for callbacks to get what they need from, e.g., `event.args` and `event.kwargs`. Is there a reason this approach doesn't work for your use case?

KarolOlko:
Hi, thanks for your reply. Well, it would work, but now all callbacks need an unused ```event``` (or ```*args``` or whatever) function parameter in their signature. I think client code is cleaner if you don't impose this limitation. If I don't need event data and base my callback only on internals - library should respect that.

PR #221, opened 2017-05-16
Argument matching for callbacks added

KarolOlko:
Let me copy my previous post here. I made sure conditions are also covered by argument matching mechanism.

This was bugging me for a while.
Say I have an trigger and not all callbacks need its arguments (especially conditions, like):
```
class Model(object):
  states = ['unbroken', 'broken']
  def __init__(self, is_concrete=True):
    self.is_concrete = is_concrete
    self.machine = Machine(model=self, (...))
    self.machine.add_transition('hit_hammer', 'unbroken', 'broken',
           conditions='was_hit_hard', unless='is_concrete')
  def was_hit_hard(self, force):
    return force > 50
  def is_concrete(self):
    return self.is_concrete

model.hit_hammer(force=50)
```
Now the code crashes and I had to resort to changing callback signature to take (*args, **kwargs), but I would like to avoid that.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -406,6 +406,59 @@ class TestTransitions(TestCase):
         self.assertTrue(m.before_state_change[0].called)
         self.assertTrue(m.after_state_change[0].called)
 
+    def test_machine_matches_event_data_to_callback_signature(self):
+        class TestModel(object):
+            def __init__(self):
+                self.b_mock = MagicMock()
+                self.c_mock = MagicMock()
+
+            def false_condition(self):
+                return False
+            
+            def on_enter_B(self):
+                self.b_mock()
+
+            def on_enter_C(self):
+                self.c_mock()
+            
+            def on_enter_D(self, a, b):
+                self.a = a
+                self.b = b
+
+            def on_enter_E(self, *args, **kwargs):
+                self.args = args
+                self.kwargs = kwargs
+
+        model = TestModel()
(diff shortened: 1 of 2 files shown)


aleneum:
Hello @KarolOlko,

first, let me say thank you for your input. We strive for improving transitions wherever suitable.
The struggle between convenience, flexibility and code base clarity has been a constant factor so far. 

This being said I -- personally -- do not feel like the balance of added value compared to added code complexity is held here. I also experienced the need of having `*args` and `**kwargs` in callbacks as a mild inconvenience. But I fear that the effort saved here has to be spent elsewhere when mapping goes wrong. Basically your suggestions reimplements one of the core functionalities of Python which is dynamic parameter mapping. Dynamic introspection is an amazing tool but doing it in every callback and check (which comes close to everywhere for everything) is a big point of failure. Especially because Python has many variants of *callable* objects and `inspect` itself is a moving target. 

I just peeked at issues with `inspect` and found that a) `inspect.getarcspec` is [deprecated](https://docs.python.org/3/library/inspect.html#inspect.getargspec) in Python 3.x and the replacement `inspect.signature` [does not handle partials well](https://bugs.python.org/issue30149). We use partials quite a lot.

tl;dr: Imho this may requires more effort for maintenance of the library as well as user code in the long run than it saves.

aleneum:
Closing this due to previously discussed maintenance and stability issues. Feel free to open it again in case you like to discuss this further.

PR #222, opened 2017-06-02
Don't set global warnings module state

ksandeep:
I am trying to use transitions for an existing django project. The warnings.simplefilter('default') line is modifying the state of warnings module globally and causing logspew.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -14,7 +14,6 @@ from functools import partial
 from six import string_types
 
 import warnings
-warnings.simplefilter('default')
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


aleneum:
Hello @ksandeep,

thank you for bringing this to our attention. 

> The warnings.simplefilter('default') line is modifying the state of warnings module globally and causing logspew.

While this is definitely not intended behaviour, removing the filter altogether will also hide our `PendingDeprecationWarnings`. I just pushed a change to `master` with a more specific filter which should only show our warnings in case it is necessary. If this still clutters you log, let me know and if possible provide a minimal example or your issue that I can use to adjust the warning output.

Issue #229, opened 2017-06-27
State feature: Timeout


aleneum:
Adds the ability to timeout an entered state. See: [states.py](https://github.com/pytransitions/transitions/blob/dev-state-extensions/transitions/extensions/states.py#L35)

### Keywords

* `timeout` (int, optional) -- if passed, an entered state will timeout after `timeout` seconds
* `on_timeout` (string/callable, optional) -- will be called when timeout time has been reached

### Exceptions:
  - will raise an `AttributeError` when `timeout` is set but `on_timeout` is not

### Example Usage

```python
from transitions import Machine
from transitions.extensions.states import add_state_features, Timeout

import time


@add_state_features(Timeout)
class CustomMachine(Machine):
    pass


class Model(object):

    def timeout(self):
        print("Timeout!")
        self.to_A()

model = Model()

states = ['A',
          {'name': 'B', 'timeout': 0.3, 'on_timeout': 'timeout'},
          {'name': 'C', 'timeout': 0.9, 'on_timeout': model.timeout}]

m = CustomMachine(model, states=states, initial='A')
model.to_B()
time.sleep(1)  # >>> "Timeout!"
print model.state  # >>>  'A'
```

### Remark

This implementation is not thread-safe! The user may either lock the model or use `LockedMachine`.

Synss:
Hi, great to see this coming. Just two small things: could you put a comment to #198: I was following #198 and overlooked this thread until yesterday. I would also check the presence of `on_timeout` with [EAFP](https://docs.python.org/2/glossary.html#term-eafp)

```diff --git a/transitions/extensions/states.py b/transitions/extensions/states.py
index 40e9ce7..1d8a01f 100644
--- a/transitions/extensions/states.py
+++ b/transitions/extensions/states.py
@@ -36,8 +36,9 @@ class Timeout(object):
     def __init__(self, *args, **kwargs):
         self.timeout = kwargs.pop('timeout', 0)
         if self.timeout > 0:
-            self.on_timeout = kwargs.pop('on_timeout', None)
-            if self.on_timeout is None:
+            try:
+                self.on_timeout = kwargs["on_timeout"]
+            except KeyError as exc:
                 raise AttributeError("Timeout state requires 'on_timeout' when timeout is set.")
         self.runner = {}
         super(Timeout, self).__init__(*args, **kwargs)
```

This way, you avoid the ``if`` in the normal case.

aleneum:
Hi @Synss, 

[...]

PR #235, opened 2017-07-12
let Machine.add_states() take **kwargs

Synss:
This can be useful for state mixins, see also Issue #229.

—as discussed.

Tested with:

```
from transitions import Machine
from transitions.extensions.states import add_state_features, Timeout

import time


@add_state_features(Timeout)
class CustomMachine(Machine):
    pass


class Model(object):

    def timeout(self):
        print("Timeout!")
        self.to_A()

model = Model()

states = ['A',
          {'name': 'B', 'timeout': 0.3, 'on_timeout': 'timeout'},
          {'name': 'C', 'timeout': 0.9, 'on_timeout': model.timeout}]

m = CustomMachine(model, states=states, initial='A')
model.to_B()
time.sleep(1)  # >>> "Timeout!"
print model.state  # >>>  'A'


def hello():
    print("hello")


m.add_state("bob", timeout=0.5, on_timeout=hello)
model.to_bob()
time.sleep(1)

```

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -584,7 +584,7 @@ class Machine(object):
         self.add_states(*args, **kwargs)
 
     def add_states(self, states, on_enter=None, on_exit=None,
-                   ignore_invalid_triggers=None):
+                   ignore_invalid_triggers=None, **kwargs):
         """ Add new state(s).
         Args:
             state (list, string, dict, or State): a list, a State instance, the
@@ -614,7 +614,7 @@ class Machine(object):
             if isinstance(state, string_types):
                 state = self._create_state(
                     state, on_enter=on_enter, on_exit=on_exit,
-                    ignore_invalid_triggers=ignore)
+                    ignore_invalid_triggers=ignore, **kwargs)
             elif isinstance(state, dict):
                 if 'ignore_invalid_triggers' not in state:
                     state['ignore_invalid_triggers'] = ignore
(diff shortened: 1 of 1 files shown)


aleneum:
weird, the build [fails](https://travis-ci.org/pytransitions/transitions/builds/252929364) for python 2.7 and python 3.3 due to a recursion errors in `dill`. I cannot reproduce this with `tox` though 😕 .

Synss:
interesting... that triggered an infinite recursion... I'll see if I can reproduce that.

aleneum:
Maybe this is a one time phenomena... Could you extend the `add_states` docstring to cover the new parameter? This will not change the actual code but a) it's a nice thing to do anyway and b) it will trick travis into building again.

aleneum:
Well... this [did not help](https://travis-ci.org/pytransitions/transitions/builds/252961932) (bummer). Maybe [this issue](https://stackoverflow.com/questions/44981303/python-dill-0-2-7-throwing-max-recursion-depth-exceeded-error) with dill 0.2.7 is related to what we experience. However, I also installed dill 0.2.7 and manually added `**kwargs` to `add_states` but everything works fine...

Synss:
Yes, I have seen that. I will see if I can reproduce this tomorrow...

aleneum:
I just rebuilt the 'state-extension' branch and (as expected) it [breaks too](https://travis-ci.org/pytransitions/transitions/builds/250291251) now ⛈ .

Issue #198, opened 2017-03-11
State timeouts?


Eric24:
It's a very interesting project with a lot of rich functionality. One of the the things I was looking for in an FSM was the ability to automatically transition to a specified state after a period of time in the current state. For example, the FMS is in "tank-filling" state, waiting for an external sensor to tell me the tank is full and move on to "tank-full" state, but if that external event doesn't happen within 30 seconds, I instead want to move to "tank-fill-error" state. Of course I could implement something like this with an external timer that fires after 30 seconds and triggers a transition from "tank-filling" to "tank-fill-error", but it seems like the use-case for this sort of thing is common enough to be built in--essentially, for any state transition, the ability to pass an optional "timeout state" and timeout value. I'd love to hear your thoughts on this.

aleneum:
Hi @Eric24,

I also use timeouts in my projects but in my oppinion this should be handled by the model or the specific state. Let's see if others may have another view on this.

Until then, a straight forward way to do it in the model could look like this:

```python
from threading import Thread
import time
from transitions import Machine


class Timeout(Thread):

    def __init__(self, func, timeout):
        super(Timeout, self).__init__()
        self.func = func
        self.timeout = timeout
        self.canceled = False
        self.start()

    def run(self):
        time.sleep(self.timeout/1000.0)
        if not self.canceled:
            self.func()


class Model:

    def __init__(self):
        self.timer = None

    def on_enter_B(self):
        self.timer = Timeout(self.timeout, 1000)

    def on_exit_B(self):
        self.timer.canceled = True


model = Model()

transitions = [['timeout', 'B', 'C']]
machine = Machine(model, states=['A', 'B', 'C'], transitions=transitions, initial='A')
model.to_B()
print(model.state)  # >>> B
time.sleep(1.1)
print(model.state)  # >>> C
```

or as part of a subclassed `State` which allows easier individual timeout handling:

```python
class TimeoutState(State):

    def __init__(self, timeout):
        self.timeout = timeout

b = TimeoutState('B', enter='set_timeout')
m = Machine(model, states=['A', b, 'C']) ...
```

Eric24:
Yes, that's essentially my current approach. I'm curious as to why you feel that this shouldn't be part of the state machine "internals"? To me, it seems like "core functionality", and as such, why trouble the developer to handle their own timeouts explicitly? For my purposes, having 30 or more states, most with timeouts, it seems like a lot of unnecessary boilerplate.

aleneum:
[...]

PR #237, opened 2017-07-14
Add `Timeout.set_timeout()` method

Synss:
In most cases, `transitions` allows to `machine.get_state(state).add_callback(function)`. However, with `Timeout`,
 - `add_callback` would have to take extra `*args`,
 - `Timeout.on_timeout` is not a list of callbacks but a single function so that `add_callback` would be misleading (the user is not *adding* anything but either setting a callback or replacing the one in place)

I would therefore propose to add a `set_timeout(self, timeout, on_timeout)` method for consistency.

### DIFF
--- a/transitions/extensions/states.py
+++ b/transitions/extensions/states.py
@@ -58,6 +58,10 @@ class Timeout(object):
             t.cancel()
         super(Timeout, self).exit(event_data)
 
+    def set_timeout(self, timeout, on_timeout):
+        self.timeout = timeout
+        self.on_timeout = on_timeout
+
 
 class Volatile(object):
 


Synss:
As far as I understand, transitions must be instantaneous, otherwise the machine would not be in any state during the transition. If you exit a state and wait for a timeout before entering the next state, this waiting-for-a-timeout period must be another state that can be represented with this extension.

Or am I completely missing the point?

aleneum:
> In most cases, transitions allows to machine.get_state(state).add_callback(function)

The purpose of `State.add_callback` is to add callbacks for entering and exiting states and takes two arguments. See the docstring:

```python
    def add_callback(self, trigger, func):
        """ Add a new enter or exit callback.
        Args:
            trigger (string): The type of triggering event. Must be one of
                'enter' or 'exit'.
```

If I get your suggestion right, you would like to the for `Timeout.add_callback` to handle not just 'enter' or 'exit' but also allow 'timeout'. We could just listify `on_timeout` and override `add_callback` to allow `machine.get_state(state).add_callback('timeout', function)`. However, extending the convenience function `Machine.on_enter/exit_<statetname>` to also allow `Machine.on_timeout_<statetname>` would require changes in `core` (which I try to avoid).

> I would therefore propose to add a set_timeout(self, timeout, on_timeout) method for 

If I am not mistaken (which may be the case) getters and setters are not really pythonic as `Timeout` properties could be altered directly. If we go for the `on_timeout` listify solution, I'd rather use property decorators to make sure that timeout callbacks are always a list.

=> I will alter `Timeout` in a way that allows to handle multiple on_timeout callbacks and extend the add_callback method accordingly. If this does not achieve what you had in mind, feel free to give feedback.

Synss:
> I will alter Timeout in a way that allows to handle multiple on_timeout callbacks and extend the add_callback method accordingly.

I am not sure this is a good idea, in my case at least, I use `on_timeout` to transition to either `next_state` or some error state. I cannot really think of any other meaningful thing to do in `on_timeout` actually. If this is correct, you should really *not* listify `on_timeout`.

Accessors are not pythonic. But in the present case, the user may want to set two attributes at once.

It is true that you can `machine.get_state(name).timeout = 10.0` and then `machine.get_state(name).on_timeout = "next_state"`. Maybe this is good enough.

aleneum:
> Generally, timeouts are probably better served ... as custom transitions, that only trigger if one lingers 

The motivation for timeout states originates from issues as stated in #198 and similar requests on Stackoverflow. I thought of timeout states and timeout transitions in a way that both fulfil different purposes: The first can be used if a state should not be *active* longer than a certain amount of time and the second if a transition *attempt* takes too long.

You suggest a mixture of both concepts (if I got this right) where when a state is entered, outgoing transitions are evaluated for 'timeout' conditions:

```python
transition = [{'trigger': 'slow_timeout', 'timeout': 100, 'src': '*', 'dest': 'Error'},
                     {'trigger': 'fast_timeout', 'timeout': 10, 'src': 'TimeCritical', 'dest': 'Error'}]
```

I can see this work too. The decorating process however becomes more complicated since it requires extended functionality for transitions and states. If you have a suggestion how this could be handled by `transitions` I am all ears.

> As far as I understand, ... machine would not be in any state during the transition

Until `on_exit` callbacks are processed, the **model** will be in the source state. The machine will switch the model's state and start to process destination callbacks. This means (thanks to Tal's design), the processed model should always be in defined state. Bringing threads into the game does make things more complicated of course (dealing with pending/ongoing transitions and such, racing conditions...) I would consider it good practise to do the actions that might timeout in preparation and condition checks

aleneum:
> I use on_timeout to transition to either next_state or some error state
I would also think that for most use cases a single callback is enough. Does this mean you (plan to) change `timeout` and `on_timeout` values during runtime? If so, what is your motivation?

> If this is correct, you should really not listify on_timeout.

well, it does not hurt 😄 . We do the same handling for `before`,`prepare` and such. As a module user you can treat all of these properties as single entities *or* lists, depending on your needs.

tyarkoni:
> However, extending the convenience function Machine.on_enter/exit_<statetname> to also allow Machine.on_timeout_<statetname> would require changes in core (which I try to avoid).

Haven't thought through this in detail, but couldn't we monkey patch `State` when `add_state_features`  in the new [`states` extension module](https://github.com/pytransitions/transitions/blob/dev-state-extensions/transitions/extensions/states.py) is invoked? Admittedly not the most elegant solution, but it could provide the desired functionality without cluttering up core.

If we think there might be other decorator-based extensions to `State` that might need to dynamically add methods, we could probably create a base `StateExtension` class all the other classes inherit from; then, in `__init__`, we would look for a `_dynamic_methods` class property, and, if found, would monkey-patch each of the listed methods into the `State` class. Would something like that work?

aleneum:
Hi @tyarkoni,

currently, callbacks are properties of `Machine`. This means `Machine` had to be adapted to alter the behaviour of `Machine.on_enter/exit/<dynamic_method>`. Since all callbacks are either a) relate to states or b) relate to transitions it might be a good idea to couple these callback keywords to `Transition` and `State` instead.

```python
# change this... in Machine._identify_callback
callback_type = cls.callbacks[[name.find(x) for x in cls.callbacks].index(0)]
# into this...
for cb in itertools.chain(cls.state_cls._dynamic_methods,
                          cls.transition_cls._dynamic_methods):
    if name.find(cb) > -1:
       callback_type = cb
       break
else:
    return None, None
```

`add_state_features` creates a class called `CustomState` from all passed Mixins and the appropriate `State` class (depending on if the a `Machine`, `LockedMachine` or `HierarchicalMachine` has to be extended). We could monkey patch `_dynamic_methods` something like this:

```python
CustomState._dynamic_methods = set(itertools.chain([c._dynamic_methods for c in inspect.getmro(CustomState)]))
```

Does this comply to what you have in mind?

PR #239, opened 2017-07-23
Fix graph crash with partial conditions

Synss:
Graph.rep() assumes that every callable has a `__name__` attribute.
However, callables like `functools.partial` and other classes with a
`__call__()` method have no `__name__`.  This patch modifies
`Graph.rep()` to first check whether the condition is a string and
return it, then try to return `f.__name__` (that is the current
behavior) and if that fails, then it calls `str()` on the argument.

`str()` should never fail and it is now the responsibility of the user
to return something meaningful from `__str__()`.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -5,6 +5,7 @@ except ImportError:
 
 from .utils import InheritedStuff
 from .utils import Stuff
+from functools import partial
 import sys
 from transitions import Machine, MachineError, State, EventData
 from transitions.core import listify, prep_ordered_arg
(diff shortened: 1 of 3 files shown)


Synss:
If you like, I can explicitly handle `functools.partial` and return `"check(True)"` for

```
def check(result):
    return result

... conditions=partial(check, True)...
```

aleneum:
Partials for condition checks haven't been tested anywhere else yet. If you call `fly` and make sure that both conditions are checked this would also test if partials are handled correctly. Just in case someone introduces something later on which breaks this compatibility.  Like:

```python
# [...]
m = self.machine_cls(states=['A', 'B', 'C'], initial= 'A', show_conditions=True,
                     title='a test')
m.add_state({'name': 'E'})
m.add_transition(trigger='fly', source='A', dest='B',
                 conditions=Check(False))
m.add_transition(trigger='fly', source='A', dest='C',
                 unless=functools.partial(check, False))
m.fly()
self.assertTrue(m.is_C())
# [...]
```

Synss:
I could test partial conditions in `test_core`, make `Graph.rep()` a `staticmethod` (it does not use `self` anyway) and test `Graph.rep()` directly. Is that OK?

aleneum:
> I could test partial conditions in test_core, make Graph.rep() a staticmethod (it does not use self anyway) and test Graph.rep() directly. Is that OK?

Sounds good. Currently, the 'coverage problem' is, that `test_anonymous_callable_conditions` does not actually makes use of `Check` and the partial function `check` though (see [coverage report](https://coveralls.io/builds/12512565/source?filename=tests%2Ftest_graphing.py#L81)). So yeah, testing partials for condition checks in `core` does make sense as well as making `rep` a module function (does not even have to be part of `Graph`)  but it does not solve the issue of unused constructs in the test itself.
I know its not a big issue but I would prefer if this could be solved in a meaningful/contributing way.

Synss:
Tests passed on my machine with 2.7.13, I'll try again...

aleneum:
Looks good to me, one function name was misspelled, thats why the test hadnt been executed.
Thanks!

PR #242, opened 2017-07-26
Include all files in the PyPI releases

jodal:
- Removed MANIFEST which is rebuilt by setuptools when needed.
- Included tests, examples, etc in MANIFEST.in.
- Added tox test environment which runs check-manifest to validate the
  MANIFEST.in file.

Downstreams like Debian prefer the PyPI packages to include everything
from the Git repo, especially changelogs and test suites. To make the
installation with pip as lightweight as possible, it is encouraged to
build and upload Python wheels in addition to the sdist tarball.


aleneum:
Sounds reasonable even though I have to admit that I lack experience when it comes to proper packaging. If @tyarkoni agrees, I will merge this pull request and have a look in how to provide wheels as well.

jodal:
Great! :-)

The following is the gist of building wheels in addition to sdist tarballs and securely uploading them to PyPI with `twine`.

```
pip install twine
python setup.py sdist bdist_wheel
twine upload dist/*
```

aleneum:
As the changes do not have any impact on how transition can be used I will treat no further feedback within 7 days as silent agreement and merge the pull request. The first release which will provided with the extended manifest and as a wheel will probably be `0.6`

tyarkoni:
Sorry, yes, sounds good to me!

PR #254, opened 2017-09-03
Add Machine.add_transitions to core

Synss:
[Note the plural]

This is consistent with `add_state` (singular) and `add_states` (plural)
and also with the ability of `__init__` to take a list of transitions.

I simply extracted the routine in `__init__` to its own method.

---

As a side note, I would prefer that `add_states` and
`add_transitions` take star-args so that the user does not need
the `[]` (easily forgotten)...

Something like:

```
...
    add_states(self, *states):
        for s in states:
            self.add_state(s)

    add_transitions(self, *transitions):
        for t in transitions:
            self.add_transition(t)
...

machine.add_states(STATE1, STATE2)   # That seems more natural to me.
machine.add_states(*STATES)

```

I am not holding my breath, though, because that would be a
breaking change.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -440,12 +440,7 @@ class Machine(object):
             self.initial = initial
 
         if transitions is not None:
-            transitions = listify(transitions)
-            for t in transitions:
-                if isinstance(t, list):
-                    self.add_transition(*t)
-                else:
-                    self.add_transition(**t)
+            self.add_transitions(transitions)
 
         if ordered_transitions:
             self.add_ordered_transitions()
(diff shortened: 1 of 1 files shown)


aleneum:
Change looks good to me. Merged.

About that:

> As a side note, I would prefer that add_states and
add_transitions take star-args so that the user does not need
the [] (easily forgotten)...

Currently `add_states` allows to pass keywords such as `on_enter/exit`, `ignore_invalid_trigger` and custom keywords for state extensions. With `add_states(self, *states)` we lose that ability, don't we?

PR #257, opened 2017-09-27
state is called <preparing>, lowercase

fsismondi:
Fixes small typo on [ adding-features-to-states example](https://github.com/pytransitions/transitions#adding-features-to-states) 

Else when trying to execute the example you would AssertionError on

assert hero.state == 'preparing'  # Preparing for the night shift


aleneum:
thanks!

PR #266, opened 2017-11-19
Get transitions

Synss:
Hi,

This is a new `get_transitions(trigger, source="*" dest="*")`
method on `Machine`. It returns the generator of
corresponding transitions.

My use case for it is that I declare a prototype FSM in a base
class and sometimes need to add before/after callbacks
afterwards.

Moreover, I think that since `Machine` already has a few
`add/get` pairs, a `get_transitions()` method is not surprising
in the API.

Best,
Mathias

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -863,6 +863,16 @@ class TestTransitions(TestCase):
         model.blocker = True
         self.assertTrue(model.next_state())
 
+    def test_get_transitions(self):
+        self.stuff.machine.add_transition('go', ['A', 'B', 'C'], 'D')
+        self.stuff.machine.add_transition('walk', 'A', 'B')
+        go_transitions = list(self.stuff.machine.get_transitions("go"))
+        walk_transitions = list(self.stuff.machine.get_transitions("walk"))
+        self.assertEqual(len(go_transitions), 3)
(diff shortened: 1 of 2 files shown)


tyarkoni:
I think this is a reasonable feature to add, but one thing I'm not clear on is why the `trigger` argument is mandatory. Wouldn't it make sense to also make it optional, so that the user can, e.g., retrieve all transitions that have a specified source, regardless of trigger?

Synss:
@tyarkoni I guess so. However, ```trigger``` is also mandatory to ```remove_transition``` and I just reuse the signature. What do you think?

tyarkoni:
I don't think consistency matters much here... removing transitions feels to me like a less common operation, and since it's destructive, it's not crazy to make removal harder to do accidentally.

Oh, also: let's return a list rather than a generator. This will almost never be a memory-intensive operation, and for the sake of user-friendliness and consistency with the rest of the API, we should avoid returning a generator when the user isn't going to necessarily being expecting one.

Synss:
OK, I'll work on that.

Synss:
I am done. You can see if that is ok now.

aleneum:
Thank you (again) for your contribution @Synss 👍

PR #274, opened 2017-12-18
fixed HierarchicalMachine.add_model() not passing arg to super

illes:


aleneum:
Hi @illes,

thank you for the report. I guess we have to rework the testing of the related methods if it could not track this bug.

PR #283, opened 2018-01-14
Update header typo in README.md

MSumulong:
Fixed "Reflexive transitions from multiple states" header


aleneum:
Thank you @MSumulong

PR #289, opened 2018-02-12
Add support for string as path to callables

ollamh:
# Problem

Right now to define a callback I need to pass a callable or string attribute of a model. This requires lots of imports to be done and can lead to circular import errors when dealing with complex projects.

# Solution
To add as an option a path to a certain function in the format `path.to.module.function_name`. 
So, the workflow changes to:
1. If the func is callable, just call it
2. If this is a string, we will try to import string assuming it is a path to a func
3. Fallback to a model attribute


aleneum:
Hi @ollamh, 

sorry for the lack of feedback. First of, I think having the ability to address callables by their module path is a valuable addition to transitions. I use a comparable approach myself and agree that this adds flexibility. The reason why I am hesitating is because of the way the different callable cases are processed. The proposed 'standard' case of using the names of model functions will require catching exceptions (`ImportError`) every time a callback is executed. Before I merge this I want to evaluate the performance impact this has and might also try to do something about it. I can't give you a reliable estimate about when this will happen but its definitely on my list.

ollamh:
@aleneum Sorry to disturb you, but is there any update on this one?

aleneum:
Hi @ollamh, no updates so far, sorry. Thanks for the reminder. It should be done mid/end next week.

aleneum:
Hi @ollamh and @paulbovbel,

>  thought I'd offer an unsolicited outside perspective

thanks for that! I always appreciate having productive feedback such as yours!
I agree that having the function to resolve callbacks decoupled from the processing context increases the flexibility of the library.
I refactored this pull request as seen in the branch [pr-289](https://github.com/pytransitions/transitions/tree/pr-289).
@ollamh: Please have a [look](https://github.com/pytransitions/transitions/compare/e660098...pr-289) if this still fulfils your needs.

I decide to use `Machine.resolve_callback(func, event_data)` for this since it eases the customisation for library users. This may help to keep state and transition definition as readable as possible.
I also decided to skip the callable test. In the past we have only checked for `string_types` and assumed the passed `func` to be callable otherwise.
Additionally, I moved the model attribute check to be conducted before the import approach since I think it is currently way more common.
A small test with a model using two attributes and one import function shows a significant performance benefit when less exceptions have to be caught.
The overall footprint of resolving callbacks is relatively small though.

Looking forward to your thoughts

ollamh:
@aleneum Look good! Actually it is more clean solution.

PR #299, opened 2018-05-11
Support internal transition

maueki:
With this patch, if the transition's dest is None, the transition will be an internal transition.
With internal transition, no state change occurs.


aleneum:
Hello @maueki,

thanks for the input. The library already supports reflexive transitions (exit and enter the same state). Could you give an example where this behaviour is not sufficient but actually no transition at all should happen?

maueki:
Hi @aleneum ,

e.g. https://en.wikipedia.org/wiki/UML_state_machine#Internal_transitions

PR #300, opened 2018-05-16
Fix missing () in extensions/states

Synss:
[Thread.is_alive()](https://docs.python.org/3/library/threading.html#threading.Thread.is_alive) is a function and the call operator `()` is missing line 106 of `transitions/extensions/states.py`.

This is obviously a bug and here is the fix!


aleneum:
Hi @Synss,

thanks for the fix!

PR #301, opened 2018-05-25
Fix installation on Python 3 with non-UTF-8 locale

jodal:
When `setup.py` is executed, it reads `README.md` using the systems default encoding. However, `README.md` contains UTF-8 encoded characters.

If installing transitions under Python 3 on a system with a non-UTF-8 locale, like `C`, Python crashes with an `UnicodeDecodeError` because it fails to decode `README.md` as ASCII, which is the default encoding in the C locale.

To reproduce:

```
virtualenv -p /usr/bin/python3 venv
. venv/bin/activate
LANG=C LC_ALL=C pip install transitions
```

This fix decodes `README.md` explicitly as UTF-8 in a way that works on both Python 2 and 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,11 @@
+import codecs
 import sys
 from setuptools import setup, find_packages
 
 with open('transitions/version.py') as f:
     exec(f.read())
 
-with open("README.md") as f:
+with codecs.open('README.md', 'r', 'utf-8') as f:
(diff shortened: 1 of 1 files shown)


aleneum:
Hello @jodal, thank you for that fix. I will create a new release asap.

aleneum:
The release `0.6.8` has been published. I tested the installation from source with `pip` and it seems to work now. Thanks again for the report.

PR #328, opened 2018-12-04
add space after machine name in log messages

cglewis:
makes the logs more readable


cglewis:
ah, ok. that makes sense. I was using the `BASIC_FORMAT`

PR #337, opened 2019-02-18
Run tests on Python 3.7 on Travis

jodal:
Since Python 3.7 was taken out of the test matrix in 9dedc70517a7f30e51e2f5b86c5a26835a03d759, Travis has shipped support for Ubuntu 16.04 and with that support for running tests on Python 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 python:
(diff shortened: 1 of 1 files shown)


aleneum:
Hi @jodal,

thanks for the update!

PR #352, opened 2019-06-05
class Transition add a arg condition_args

suparek:


aleneum:
Hello @suparek,

sorry for the long wait. As far as I can tell you are interested in two features:

1. collection all results of enter and exit state callbacks
2. assigning parameters to transitions which are used in the transition condition checks (and prepare/before/after callbacks). The general idea is probably to keep the condition check more generic and let parameters be part of the json/yaml/etc. machine configuration

Feature 2. sounds interesting and I will definitely give it a thought.
I will close this PR for now though as it is broken and also stale. If you consider continue working on it please make sure that you test your contribution thoroughly. Feel free to comment or open a new issue if you want to discuss any of that feature in more detail.

PR #368, opened 2019-09-03
Fix setup

artofhuman:
Problem
---
assert in setup.py fails with error
```
Traceback (most recent call last):
  File "setup.py", line 11, in <module>
    assert long_description[:3] == '## '  # Description should start with a headline (## Quickstart)
AssertionError
```

```
-> assert long_description[:3] == '## '  # Description should start with a headline (## Quickstart)
(Pdb) long_description[:3]
'\n##'
```
Solution
---
Use regexp to remove text before `## Quickstart` section

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,9 +6,13 @@ with open('transitions/version.py') as f:
     exec(f.read())
 
 with codecs.open('README.md', 'r', 'utf-8') as f:
+    import re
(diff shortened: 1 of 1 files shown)


aleneum:
Yeah, stripping everything before '## Quickstart' is a way better approach than guessing how many characters have to be stripped away. Thank you for another contribution, @artofhuman!

PR #370, opened 2019-09-28
Fix: callbacks order

tzoiker:
Order of callbacks is not respected due to use of `asyncio.gather`, thus, `after_state_change`/`before_state_change` may execute before/after any `after`/`before` callback.

### DIFF
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -104,3 +104,33 @@ class TestAsync(TestTransitions):
 
         assert m1.is_C()
         assert m2.is_C()
+
+
+async def test_callback_order():
+
+    finished = []
+
+    class Machine(AsyncMachine):
+        async def before(self):
+            await asyncio.sleep(0.1)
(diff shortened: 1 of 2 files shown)


aleneum:
Hi @tzoiker,

you are right. That shouldn't happen. Thanks for reporting and fixing!

vishes-shell:
Hi, @aleneum 

This has been merged to `dev-async`, however there is no such branch anymore, and this has not been released, hence bug still remains and order is still invalid, which means completely unreliable.

How can we resolve this? Creating new issue, or creating MR?

Thanks

aleneum:
Hello @vishes-shell,

you are totally right. I have absolutely no clue why this didnt make it into `master` when `dev-async` was merged. The tests look a bit off but anyway. I merged theses changes into `master` and starting with 64e1a6d things should work (again). Thank you for finding this and pointing it out.

PR #373, opened 2019-10-21
Add support for StrEnum

chrisjbremner:
`Enum` allows you to subclass it with additional mixins to create custom `Enum`s that act in a similar fashion to what you are mixing in. For example, you can create an [`IntEnum`](https://docs.python.org/3/library/enum.html#intenum) which allows the enum to be used like an `int`. Similarly, you can make a [`StrEnum`](https://docs.python.org/3/library/enum.html#others) to do string comparisons.

Using a `StrEnum` almost works with the existing code base, but there's one small if statement that incorrectly catches `StrEnum`s but not regular `Enum`s, which I fix in this PR. I also included a test that would fail without this code change, but succeeds after.

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -105,6 +105,22 @@ class TestEnumsAsStates(TestCase):
         assert s.is_ONE()
         assert s.message == 'Goodbye'
 
+    def test_str_enum(self):
+        class States(str, enum.Enum):
+            ONE = "one"
+            TWO = "two"
+
+        class Stuff(object):
+            def __init__(self, machine_cls):
+                self.state = None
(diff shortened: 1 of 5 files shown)


chrisjbremner:
@aleneum I also found that the diagramming extensions did not work when using `Enum`s for states (because it sometimes referred to each state by the Enum's name, and other times by just the Enum itself), so I attempted to fix that as well in my most recent commit.

chrisjbremner:
Looks like my second commit broke some tests, I will fix that when I get a chance.

chrisjbremner:
Closing, as this was fixed in https://github.com/pytransitions/transitions/pull/357

Issue #377, opened 2019-10-30
Trigger methods mask away KeyError exceptions


facundofc:
If any callback raises a KeyError exception it is translated into an AttributeError with the "Do not know event named TRIGGER_NAME" message.

This is misleading and I think not the correct behavior.

aleneum:
Resolved in #378

PR #378, opened 2019-10-30
Don't mask away KeyError when executing a transition

facundofc:
This should fix issue #377.


aleneum:
Hello @facundofc,

good catch! Thank you for the PR! Wish you some nice holidays and a good start into 2020.

PR #386, opened 2019-12-25
Fix transitions for enums with str behavior (Django3)

artofhuman:
Django3 introduced enums (TextChoices), but this enums has str behavior
see https://github.com/django/django/blob/master/django/db/models/enums.py#L77

isinstance(source, string_types) returns True

To avoid this behavior, we need to remove the check on str class.

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -47,6 +47,19 @@ class TestEnumsAsStates(TestCase):
         assert m.is_YELLOW() is False
         assert m.is_GREEN() is True
 
+    def test_if_enum_has_string_behavior(self):
+        class States(str, enum.Enum):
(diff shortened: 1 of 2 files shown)


aleneum:
Hi @artofhuman,

since [this line](https://github.com/pytransitions/transitions/pull/386/files#diff-ead0bb6c6c63bccefe005346890f03a5L864) should take care of the single string case I don't see any drawback. Quite the contrary, it looks cleaner now. Thanks for the PR! Have some nice holidays and a good start into 2020!

artofhuman:
@aleneum Thanks 🎄

artofhuman:
Hi @aleneum, can we release this commit in new patch version 0.7.2 ?I tested in our project and my tests are green :)

PR #392, opened 2020-01-29
Fix add_ordered_transitions w/o initial state

mkaranki:
Previous implementation assumed that the initial state of the state
machine must be specified in the states list.

However, the add_ordered_transitions could be useful also when only a
part of the state machine is wanted to have ordered transactions. Think
for example nested state machine where sub-states of a state is wanted
to be traversed in order, perhaps in a loop.

To fix, allow calling add_ordered_transitions with states argument not
including the initial state. In effect, loop_includes_initial argument
has no relevance if the initial state is not present in the states list.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -373,6 +373,17 @@ class TestTransitions(TestCase):
         m.next_state()
         self.assertEqual(m.state, 'beginning')
 
+        m = Machine('self', states, initial='beginning')
+        # Partial state machine without the initial state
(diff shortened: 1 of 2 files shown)


aleneum:
Hello @mkaranki,

thank you for this suggestion. This feature/fix will be part of `transitions` 0.8. I implemented it a bit differently though. I allowed myself to copy your test and hope that's fine with you.

Best regards!

PR #403, opened 2020-03-25
AsyncMachine.dispatch now returns a boolean as expected

thedrow:
In https://github.com/pytransitions/transitions/blob/37fd5a31a7ba544701799bcde073f2193c033e0a/transitions/core.py#L1030-L1039 the `dispatch` method returns a boolean.
The AsyncMachine implementation did not return a value.

This PR ensures it does.


aleneum:
Hi @thedrow,

good catch! Thank you for the PR.

Issue #400, opened 2020-03-25
AsyncGraphMachine cannot be initialized with Enums as states in 0.8.0


kbinpgh:
The following snippet of code throws an error when trying to initialize an `AsyncGraphMachine`

```
from transitions.extensions import MachineFactory
from enum import Enum, auto
import time

class State(Enum):
    START = auto(),
    ONE = auto(),
    TWO = auto(),
    THREE = auto(),

def before(which=None, future=None):
    global start_time
    print(f"Before transition{which}")
    if which == 'one':
        start_time = time.time()
    
def after(which=None):
    global start_time
    print(f"After transition{which}")
    if which == 'reset':
        end_time = time.time()
        print(f"Done at {end_time} in {end_time-start_time}s")    
    
transition_1 = dict(
    trigger='one',
    source=State.START,
    dest=State.ONE,
    before=before,
    after=after,
)

transition_2 = dict(
    trigger='two',
    source=State.ONE,
    dest=State.TWO,
    before=before,
    after=after,
)

transition_3 = dict(
    trigger='three',
    source=State.TWO,
    dest=State.THREE,
    before=before,
    after=after,
)

transition_0 = dict(
    trigger='reset',
    source=State.THREE,
    dest=State.START,
    before=before,
    after=after,
)

class Model(object):
    pass

global start_time
start_time = None
model = Model()
        
machine = MachineFactory.get_predefined(graph=True, asyncio=True)(
    model=model,
    states = [v for v in State],
    transitions = [transition_0, transition_1, transition_2, transition_3],
    queued = True,
    initial = State.START,
)
```

If the line: `initial = State.START,` is commented out, there is no error thrown, but a default, unconnected inital state is created in the machine. Transitions seem to work normally with this workaround.

*Error details:*
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/.local/lib/python3.7/site-packages/pygraphviz/agraph.py in __new__(self, graph, name, nh)
   1614             try:
-> 1615                 nh = gv.agnode(graph.handle, n.encode(graph.encoding), _Action.find)
   1616             except KeyError:

KeyError: 'agnode: no key'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-58-83ccb3c51131> in <module>
     66     transitions = [transition_0, transition_1, transition_2, transition_3],
     67     queued = True,
---> 68     initial = State.START
     69 )

[...]

PR #405, opened 2020-03-25
Nested enums now translate to nested states

### LINKED ISSUE #404: Support for nested states from nested Enums
Consider the following Enum:
```python
from enum import Enum, auto
class Foo(Enum):
  A = auto()
  B = auto()

class Bar(Enum):
  FOO = Foo
  C = auto()
```

If we were to provide `Bar` to a `HierarchicalMachine` I'd expect it to take all the items in Foo as children of the Foo state.

However this is not the case:
```
In[18]: from transitions.extensions import HierarchicalMachine
In[19]: HierarchicalMachine(states=Bar)
Out[19]: <transitions.extensions.nesting.HierarchicalMachine at 0x7f517ea75d90>
In[20]: machine = HierarchicalMachine(states=Bar)
In[21]: machine.states
Out[21]: 
OrderedDict([('FOO', <NestedState('FOO')@139987994034720>),
             ('C', <NestedState('C')@139987994035296>),
             ('initial', <NestedState('initial')@139987994035200>)])
In[22]: machine.states['FOO']
Out[22]: <NestedState('FOO')@139987994034720>
In[23]: machine.states['FOO'].states
Out[23]: OrderedDict()
```

This can be easily added if I am not wrong.

thedrow:
Fixes #404.

This needs documentation but I don't know which version this feature will released on.
I'm assuming 0.9.0. Am I correct?

### DIFF
--- a/tests/test_nesting.py
+++ b/tests/test_nesting.py
@@ -244,9 +244,13 @@ class TestTransitions(TestsCore):
             FOO = Foo
             C = 2
 
-        m = self.stuff.machine_cls(states=Bar, initial='A')
+        m = self.stuff.machine_cls(states=Bar, initial=Bar.C)
 
         self.assertEqual(sorted(m.states['FOO'].states.keys()), ['A', 'B'])
+
+        m.to_FOO_A()
+        self.assertFalse(m.is_C())
+        self.assertTrue(m.is_FOO_A())
 
     def test_enter_exit_nested_state(self):
         State = self.state_cls
--- a/tests/test_nesting.py
+++ b/tests/test_nesting.py
@@ -12,6 +12,7 @@ from os import unlink
 
 from transitions.extensions.nesting import NestedState
 from transitions.extensions import MachineFactory
+from transitions.core import Enum
 
 from unittest import skipIf
 from .test_core import TestTransitions as TestsCore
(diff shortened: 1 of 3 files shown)


aleneum:
That looks good to me. Some minor tweeks need to be done though (for instance [here](https://github.com/pytransitions/transitions/blob/070b45047a7d9c253401a39086b0f1ccc5b8dfc4/transitions/extensions/nesting.py#L28)). Did you check whether the resulting Graphviz graphs look okay? Is it possible to determine the actual state and can you use auto transitions?:

```python
from enum import Enum, auto
class Foo(Enum):
  A = auto()  # 1 
  B = auto()  # 2

class Bar(Enum):
  FOO = Foo
  C = auto()  # 1

m = HierarchicalMachine(states=Bar, initial=Bar.C)
m.to_FOO_A()  # does this work?
assert m.is_C() is False  # a check by state.value would return True though 
```

thedrow:
Do we need to remove this check?
https://github.com/pytransitions/transitions/blob/070b45047a7d9c253401a39086b0f1ccc5b8dfc4/transitions/extensions/nesting.py#L27-L28

thedrow:
Setting the initial to `Bar.C` uncovered a bug in this PR that causes the test to fail because of the Graphvis extension.
I fixed it now. I'll generate a graph and see if it looks correct.

thedrow:
Looks right to me :)
g

aleneum:
Merged! Thanks again for your valuable contributions!

> Setting the initial to Bar.C uncovered a bug in this PR that causes the test to fail because of the Graphvis extension.

this is comparable to #400 and should be fixed with 0a66a36. An additional test in `set_node_style` should not be necessary anymore. I removed it.

Sidenote: Enums with identical values cannot be distinguished. That should be at least mentioned in the documentation.

thedrow:
> Do we need to remove this check?
> 
> https://github.com/pytransitions/transitions/blob/070b45047a7d9c253401a39086b0f1ccc5b8dfc4/transitions/extensions/nesting.py#L27-L28

Wait, what about this?

PR #406, opened 2020-03-25
Added coverage for duplicate states in HierarchicalMachine

thedrow:
Coverage for these conditions were missing so I added them :smiley:.

### DIFF
--- a/tests/test_nesting.py
+++ b/tests/test_nesting.py
@@ -12,6 +12,7 @@ from os import unlink
 
 from transitions.extensions.nesting import NestedState
 from transitions.extensions import MachineFactory
+from transitions.core import Enum
 
 from unittest import skipIf
 from .test_core import TestTransitions as TestsCore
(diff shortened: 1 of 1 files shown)


aleneum:
Nice! Could you add those tests to `test_enum.py` instead (in particular [here](https://github.com/pytransitions/transitions/blob/070b45047a7d9c253401a39086b0f1ccc5b8dfc4/tests/test_enum.py#L122))? This will make sure that users of Python 2.7 without the `enum34` module can run the tests as well (since enum tests will be skipped).

I see that Travis' Python 2.7 job is not failing suprisingly. More surprisingly, the tests are actually executed instead of skipped. Maybe `enum34` is installed by default since it is no dependency of `transitions`. When I run tests locally, `test_enum` will be skipped for Python 2.7.

thedrow:
I was about to do the changes you requested today.
Do I need to do them anyway or are you going to do so?

PR #407, opened 2020-03-25
CI Improvements

thedrow:
I got tired of the tests being slow so I installed xdist.
On my machine with 8 cores it cuts down the runtime from 30 seconds to 11 seconds.

On Travis it should cut the tests runtime to ~20 seconds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+cache: pip
 python:
   - "2.7"
   - "3.5"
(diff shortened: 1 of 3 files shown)


aleneum:
The most time consuming factor is testing threading. I tried to reduce waiting/sleeping times in the past but depending on the used hardware, tests behaved quite shaky. I will have a look at `xdist`.

PR #427, opened 2020-04-12
HierarchicalMachine.to_state() now works with enumerations

thedrow:
It is sometimes useful to use `to_state()` when testing even if we don't have special separator characters.
Therefor, this method should work with enumerations as states as well.


thedrow:
This PR still lacks tests.

Issue #420, opened 2020-04-05
Enable usage of other event loops in `AsyncMachine`


thedrow:
[AnyIO](https://anyio.readthedocs.io/en/latest/) is a library which abstracts event loops.
Currently our implementation only supports AsyncIO but trio is also a popular event loop which we can support.
AnyIO allows us to support both with ease.

I can port all of the code except these lines:
https://github.com/pytransitions/transitions/blob/9237b9e0978ad3d904d4f0e7b3d82d99a6b7e76f/transitions/extensions/asyncio.py#L125-L136

I think we should be using a [CancelScope](https://anyio.readthedocs.io/en/latest/cancellation.html) but I'm not really sure how.
Any feedback is appreciated.

thedrow:
I don't think we should introduce a hard dependency.

Maybe an extra dependency?
Or a different package?

I need this because Celery will be using trio.

aleneum:
I am currently attempting to decouple asyncio calls from `AsyncMachine` to make overriding easier (see [dev-trio](https://github.com/pytransitions/transitions/tree/dev-trio))
* `asyncio.iscoroutinefunction` could be replaced with `inspect.isawaitable` with minor refactoring
* calls to `asyncio.gather` have been unified to call a `AsyncMachine.await_all` instead. This should make the usage of nurseries and stuff easier. But I need to test that. 
* cancellation of running tasks has been isolated in `AsyncMachine.switch_model_context(model)`

ToDo:
* How does trio handle cancellation? `(Nested)AsyncEvent` awaits `AsyncMachine._process` and catches `asyncio.CancelledError`.

aleneum:
I just pushed changes to `AsyncMachine` that finished the isolation of `asyncio`. `AsyncTimeout` still relies on `asyncio` though. If another async library should be used, a subclass of `AsyncMachine` needs to take care of:
* `await_all` - basically the parallel callback processing in `before`, `enter`, `exit` and such
* `process_context` - a check whether this current event trigger has been called by a currently processed event or happened in a new context
* `switch_model_context` - cancel currently processed tasks when a transition is about to happen IF these tasks are processed in another event

Based on these changes I could implement a `TrioMachine`:
```python
import trio

from transitions.extensions.asyncio import AsyncMachine


class TrioMachine(AsyncMachine):

    @staticmethod
    async def await_all(partials):
        """
        Executes callables without parameters in parallel and collects their results.
        Args:
            partials (list): A list of callable functions

        Returns:
[...]

PR #445, opened 2020-06-18
Refactored AsyncMachine to enabled trio override

aleneum:
see #420

### DIFF
--- a/transitions/extensions/asyncio.py
+++ b/transitions/extensions/asyncio.py
@@ -15,8 +15,6 @@ from .nesting import HierarchicalMachine, NestedState, NestedEvent, NestedTransi
 _LOGGER = logging.getLogger(__name__)
 _LOGGER.addHandler(logging.NullHandler())
 
-is_subtask = contextvars.ContextVar('is_subtask', default=False)
-
 
 class AsyncState(State):
     """A persistent representation of a state managed by a ``Machine``. Callback execution is done asynchronously.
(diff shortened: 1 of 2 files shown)


transitions/extensions/asyncio.py
```diff
@@ -72,14 +71,11 @@ async def check(self, event_data):
                 model attached to the current machine which is used to invoke
                 the condition.
         """
-        predicate = event_data.machine.resolve_callable(self.func, event_data)
-        if asyncio.iscoroutinefunction(predicate):
```
thedrow, line 76: You can still use `asyncio.iscoroutinefunction`.
In this case, it's better.

aleneum: You'd prefer `asyncio.iscoroutinefunction` over `inspect.isawaitable`? Could you eloborate on this? I found [this SO post](https://stackoverflow.com/questions/36076619/test-if-function-or-method-is-normal-or-asynchronous) quite insightful and especially [this answer](https://stackoverflow.com/a/36077430/1617563) gave me the impression that `isawaitable` is the most generic version for that particular use case. The documentation seems to be consistent with claims made in that post as far as I can tell. But I am not attached to `isawaitable` of course. If there are -- for instance performance-wise -- facts that make `asyncio.iscoroutinefunction` more suitable, I am all for it.

thedrow: `asyncio.iscoroutinefunction` returns True for generator-based coroutine functions decorated with @coroutine according to the documentation.

I see no harm in using it.

aleneum: I agree with you that `asyncio.iscoroutinefunction` could be used independently of the chosen event loop. I'd argue however that `inspect.isawaitable` is even better here since callbacks could also be Futures which `asyncio.iscoroutinefunction` is not accepting (see [gist](https://gist.github.com/aleneum/870255aab5094f4dfb9d32408d18cd7f)). For `transitions` it should not matter whether its an async function, a generator or a Future. Eventually, we want to know if we should await something or not.

Issue #447, opened 2020-06-28
Support enums which are already an instance of State class


chrko:
I like to use enums which itself are based on the (custom) state class of the transition machine, e.g.

```python
class States(CustomState, enum.Enum):
    STATE1 = ('custom_constructor_arg1', 'custom_constructor_arg2', …)
    ...
```

This currently breaks at https://github.com/pytransitions/transitions/blob/3a17f8fb451f22061c775f771417d8b6f3bc396c/transitions/core.py#L769 because we assume enums are also only strings, but they can be more.

It should be enough to test before if `state` is already an instance of `self.state_cls`. I have monkey patched the method and it works for me:
```python
…
            if not isinstance(state, self.state_cls):
                if isinstance(state, (string_types, Enum)):
                    state = self._create_state(
                        state, on_enter=on_enter, on_exit=on_exit,
                        ignore_invalid_triggers=ignore, **kwargs)
                elif isinstance(state, dict):
                    if 'ignore_invalid_triggers' not in state:
                        state['ignore_invalid_triggers'] = ignore
                    state = self._create_state(**state)
…
```

aleneum:
Hi @chrko,

that sounds like quite some corner case but for HSMs its basically just some reordering of the type checks. For `core` its one additional check though. I created a feature branch [feature-state-check](https://github.com/pytransitions/transitions/tree/feature-state-check) and will wait for some further feedback from the community. If there is no objection I will merge these changes into master before the next release.

chrko:
Ah, perfect. I haven't checked HSM yet. :/

The test case you added test inherits from dict. I think we should add a test case which tests explicit `class States(State, enum.Enum)` or `class States(CustomState, enum.Enum)`. I will experiment until next weekend and post a patch.

chrko:
Ah, perfect. I haven't checked HSM yet. :/

The test case you added test inherits from dict. I think we should add a test case which tests explicit `class States(State, enum.Enum)` or `class States(CustomState, enum.Enum)`. I will experiment until next weekend and post a patch.

aleneum:
> I think we should add a test case which tests explicit class States(State, enum.Enum) [...]  I will experiment until next weekend and post a patch.

I added a test for enums as well. Let me know if it is missing something.

chrko:
Would you like to open a PR? It's easier to comment code there.

[...]

PR #448, opened 2020-06-29
Feature state check

aleneum:
see #447 for details


aleneum:
The UC where Enums should also act as a state is imo a corner case. I did not came up with a solution to distinguish between Enums and State+Enums that does not require additional checks for the more conventional use cases. Feel free to comment if #447 is relevant for you.

PR #461, opened 2020-08-24
Add Retry mix-in

rgov:
Note: This involved a bit of in-browser coding, so please check for formatting issues and bugs.

This change introduces a Retry mix-in which lets you limit the number of times a state may be re-entered from itself. This is useful if a state represents an operation that may need to be retried a few times, but should eventually give up.

This works really well with the Timeout mix-in:

```
{ 'name': 'PingServer',
  'timeout': 1, 'on_timeout': 'to_PingServer',
  'retries': 3, 'on_failure': 'to_CriticalFailure',
},
```

In this example, the `PingServer` state might involve sending a `ping` packet to a server and expecting a `pong` reply, whereupon it will transition to another state. If the timeout is met, then it would re-enter the `PingServer` state, triggering a second `ping` to be sent. But on the 5th time the state is entered, it will instead transition to `CriticalFailure` without sending another `ping`.

### DIFF
--- a/transitions/extensions/states.py
+++ b/transitions/extensions/states.py
@@ -5,6 +5,7 @@
     This module contains mix ins which can be used to extend state functionality.
 """
 
+from collections import Counter
 from threading import Timer
 import logging
 import inspect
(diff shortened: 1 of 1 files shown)


cliffburdick:
Hi @rgov and @aleneum, was this feature ever merged into the main branch? I noticed that there's no workaround currently because setting a timeout and going back to your original state doesn't work.

rgov:
It is on the [dev-0.9 branch](https://github.com/pytransitions/transitions/tree/dev-0.9).

PR #467, opened 2020-09-11
docs: Fix simple typo, referes -> refers

timgates42:
There is a small typo in transitions/extensions/nesting.py.

Should read `refers` rather than `referes`.


aleneum:
👍

PR #470, opened 2020-09-12
Fix diagrams_pygraphviz.Graph.get_graph method when states are enums

termim:
The example from `README.md` for `GraphMachine` fails if machine states are enums.


aleneum:
Hello @termim,

thank you for that PR. Looks good to me.

PR #473, opened 2020-09-22
Use model_attribute in predicate method name

artofhuman:
This PR adds `model_attribute` to predicate method name if model_attribute != default value (state)

It needs when we are using multiple states in the model.
Currently `transitions` generates only single method `is_{state_name}`. And if we have the same states in the model as in the test case. ['A', 'B'] for first state and ['A', 'B'] for the second state we can't use predicate methods because the instance already has this predicate for first state

Solutuon: separate predicate methods by name `is_{method_attr}_{state_name}`

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1071,25 +1071,31 @@ class TestTransitions(TestCase):
     def test_multiple_machines_per_model(self):
         class Model:
             def __init__(self):
-                self.state_a = None
-                self.state_b = None
+                self.car_state = None
+                self.driver_state = None
 
         instance = Model()
-        machine_a = Machine(instance, states=['A', 'B'], initial='A', model_attribute='state_a')
-        machine_a.add_transition('melt', 'A', 'B')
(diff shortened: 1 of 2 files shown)


aleneum:
Hi @artofhuman,

I think this change is reasonable. I fear that it will break someones code though. Some users use `model_attribute` just to rename the attribute and may still rely on `is_<state_name>`. We should provide a ... transition... period to adjust code that potentially breaks. I also changed the order of the 'if clauses'. I was once told taught is easier to comprehend when conditions are formulated 'positively' (`ìf x == 0` rather than `x != 1`) whenever possible. Let me know what you think about 1354a65baaf4a990fac55553e1176274972a6e97. Additionally, we do have similar 'issues' with `to_<state_name>` as well. Should that also be adjusted for consistency's sake?

artofhuman:
Thanks @aleneum. I think there have same problem with auto generated transitions. `to_A`, 'to_B'. I think it aslo should use prefix

aleneum:
I feel the same way. I will add this and hopefully publish 0.8.3 in the next days.

PR #479, opened 2020-10-19
Return all nested transitions

thedrow:
Unless I'm missing something, the return statement is incorrectly indented.
If that's not the case, then the while loop is redundant and can be removed.

We should add a test to verify that this issue does not regress.


aleneum:
Hi @thedrow,

this indeed looks weird. It can be simplified even further (see ea4ed0e). I also followed your recommendation and added tests for that method.

aleneum:
Closing this since it has been resolved by ea4ed0e. Thank you for reporting, @thedrow!

Issue #475, opened 2020-10-01
AsyncMachine: Memory leak happend


h-nakai:
Version: 0.8.3 (also occurred in 0.8.2)

We've been using Machine, but we've adopted AsyncMachine to add a new API call process.
However, the use of asyncmachine caused a memory leak.

I also tried gc.collect(), but it didn't improve it.
(return value is 0 and there are no content in gc.garbage)

small sample code is included here.

```
import asyncio
from transitions.extensions.asyncio import AsyncMachine


class Model(object):
    def __init__(self):
        pass


async def async_machine() -> None:
    counter = 0
    states = ["start", "end"]
    transitions = [{
        "trigger": "next",
        "source": "start",
        "dest": "end"
    }]
    while True:
        model = Model()
        machine = AsyncMachine(model=model, states=states, initial="start", transitions=transitions, auto_transitions=False)
        await model.trigger("next")

asyncio.get_event_loop().run_until_complete(async_machine())
```

```
import asyncio
from transitions.extensions.asyncio import AsyncMachine


class Model(object):
    def __init__(self):
        pass


async def async_machine() -> None:
    counter = 0
    states = ["start", "end"]
    transitions = [{
        "trigger": "next",
        "source": "start",
        "dest": "end"
    }]
    machine = AsyncMachine(states=states, initial="start", transitions=transitions, auto_transitions=False)
    while True:
        model = Model()
        machine.add_model(model)
        await model.trigger("next")
        machine.remove_model(model)
        
asyncio.get_event_loop().run_until_complete(async_machine())

```

h-nakai:
I found.
The dict named async_task,
size of this dictionary is on the rise

https://github.com/pytransitions/transitions/blob/dd787e1ba88408e2efb523bb7bd5c3fd423b9cdb/transitions/extensions/asyncio.py#L305

https://github.com/pytransitions/transitions/blob/dd787e1ba88408e2efb523bb7bd5c3fd423b9cdb/transitions/extensions/asyncio.py#L366

aleneum:
Hello @h-nakai,

thanks for that very precice issue description. I never imagined that this could be a problem but if your use case actually involves creating that many models, I will implement a clean up for that dictionary.

h-nakai:
@aleneum 
Thanks for the reply.
We plan to create many sessions, so it would be helpful if you could make some corrections.
I appreciate your help.

Tentatively.
We'll going to override the Machine's __init__.
[...]

PR #490, opened 2020-12-09
Add transition queue for each model in AsyncMachine

jekel:
Main purpose for separate queue for each model is to make them executed in parallel, and not sequential for all triggers from all models common queue  
Otherwise if some state in one models needs many time to be processed - all others will wait, and if that model triggers state to another - that trigger will be queued and can be executed more later, after all others triggers from another models execute
thats isn't quite right
Also, in case of exception inside transition - queue is cleared not for all machine, but only for model where exception was raised

### DIFF
--- a/transitions/extensions/asyncio.py
+++ b/transitions/extensions/asyncio.py
@@ -3,7 +3,7 @@ import logging
 import asyncio
 import contextvars
 import inspect
-
+from collections import defaultdict, deque
 from functools import partial, reduce
 import copy
 
@@ -306,6 +306,10 @@ class AsyncMachine(Machine):
     protected_tasks = []
     current_context = contextvars.ContextVar('current_context', default=None)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._transition_queue = defaultdict(deque)
+
     async def dispatch(self, trigger, *args, **kwargs):  # ToDo: not tested
         """ Trigger an event on all models assigned to the machine.
         Args:
@@ -384,7 +388,7 @@ class AsyncMachine(Machine):
             else:
                 self.async_tasks[model] = [asyncio.current_task()]
             try:
-                res = await self._process(func)
+                res = await self._process(func, model)
             except asyncio.CancelledError:
                 res = False
             finally:
@@ -392,10 +396,10 @@ class AsyncMachine(Machine):
                 if len(self.async_tasks[model]) == 0:
                     del self.async_tasks[model]
         else:
-            res = await self._process(func)
+            res = await self._process(func, model)
         return res
 
-    async def _process(self, trigger):
+    async def _process(self, trigger, model):
         # default processing
         if not self.has_queue:
             if not self._transition_queue:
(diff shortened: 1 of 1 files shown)


aleneum:
Hello @jekel,

an interesting and reasonable addition. Let's say we want to keep the current behaviour as the default behaviour but enable your workflow as well. As a developer, would you prefere another parameter such as  `AsyncMachine(queued=True, parallel_queue=True)` or another value for `queued` such as `AsyncMachine(queued='parallel')`.

jekel:
Hello
I think you are right, that it would be better to make new behaviour as option `queued='model'`, and later after it will be test by more people - make it default.
I will make commit with changes

aleneum:
Hi @jekel,

> I will make commit with changes

I already prepared some adjustments and waited for your feedback. So no need for adaptions from your side. I think I will be able to merge the PR today and release a new version today or tomorrow. Again, thank you for your contribution. It's always nice when an idea/suggestion is submitted so well prepared and outlined!

jekel:
@aleneum i have another idea - when `remove_model` is called on machine, we need also to clear all queues and stop processing all transitions from removed model - what do you think about it? but it could be done only when we have separate queue for each model..

aleneum:
> when remove_model is called on machine, we need also to clear all queues and stop processing all transitions from removed model

Good point. I'll see what can be done about it and will let you know.

aleneum:
just a short update: i wrote some tests (which i didnt account for in my previous optimistic time estimation) for the 'model' queue mode and also added some documentation. removing queued events (from a single queue) is a bit more complicated than expected.I'll keep you updated.

jekel:
Yeah.. its a good question how to clean queue..
I think, that the best way is to replace it with new - filtered by rule - model class is not the model to be removed

aleneum:
The filtering itself is actually quite straight forward

```python
    def remove_model(self, model):
        models = listify(model)

        for mod in models:
            self.models.remove(mod)
            if self.has_queue == 'model':
                del self._transition_queue[mod]
            elif self.has_queue:
                self._transition_queue[mod] = deque([e for e in self._transition_queue if e.args[0] != mod])
```

but the processing in `Machine._process` does not expect the queue to be changed.

aleneum:
I merged your changes into `feature-model-queues` (see [diff](https://github.com/pytransitions/transitions/compare/master...feature-model-queues)) but had to take another approach with a `dict` rather than a `defaultdict`. In #475, empty dictionary keys (with empty lists) caused quite some memory leak when models were constantly added and removed. This is why I think explicitly adding and removing dictionary keys with `deque` or `self._transitions_queue` is a better solution. Alternatively, one could use `defaultdict` but remove a key whenever a list is empty. That, however, has probably some considerable performance impact.
Furthermore, I struggled a bit with writing useful tests for using and removing model queues.

What's your oppinion about `dict` instead of `defaultdict`? Could you think of more streamlined tests for `test_model_queue` and `test_queued_remove`?

aleneum:
hmm... removing_model still does not work as intended. wont work with more than one model left in the list and also result in multiple event queues

jekel:
So...
first of all, if there is only one way to work with queues - from inside, i'm ok that list is created and destroyed from Machine explicitly -> `dict` is fine

another thing, in your code, you assume that `remove_model` is called from inside running transition,  
[here](https://github.com/pytransitions/transitions/compare/master...feature-model-queues#diff-938de46e643ab091cff6a7c23e4e752e8907e2266f47889d988923352f7a1058R650) and [here](https://github.com/pytransitions/transitions/compare/master...feature-model-queues#diff-6886dfaede3b127f0d73b71c33a7f6b19ce43c3d65bd443d40e7bec3fe5ecf06R612)  
so i think there will be exception when `remove_model` will be called when queue is empty.

like in my case -> model is removed from outside and there can be no running transitions

aleneum:
>  so i think there will be exception when remove_model will be called when queue is empty.
right, need to add this to the tests

I implemented a `_DictionaryMock` which ignores keys and always returns `self._transitions_queue`. This way, handling of model queues and single queue can be kept similar.

aleneum:
resolved in #492

PR #506, opened 2021-02-01
Fix resolving enums when they are set as nested initial states

thedrow:
If we set a nested state's initial value to an enum it doesn't get resolved and raises a KeyError instead.
I'm not sure which test case reproduces this directly though so I'd appriciate if you can write a unit test for this.


aleneum:
Travis is unfortunately overloaded (again) but the change looks good to me.

thedrow:
How can we write a test for this? It only reproduces with deeply nested states as far as I can see.

aleneum:
I assume that something like [this](https://github.com/pytransitions/transitions/blob/b6a1d15bd19738b7790136a66ee71611ccff82a1/tests/test_parallel.py#L151) can be rewritten to use `Enum`s instead.

PR #531, opened 2021-04-12
`listify()` now supports empty weak references

thedrow:
If you attempt to listify an empty weak reference you get a `ReferenceError` since comparing their type to something is no longer possible.
Instead, just return a list containing the object.


thedrow:
If you're wondering why this is important, this happens when you try to remove a model which is a weakref to an actual object.
There are more problems but this solves the most basic one.

PR #538, opened 2021-06-24
setup.cfg: change description-file to description_file

ConiKost:
UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead.

Signed-off-by: Conrad Kostecki <conikost@gentoo.org>


aleneum:
Thanks @ConiKost!

Issue #541, opened 2021-08-21
README.md - Missing "Machine" import


kauedg:
The main `README.md` file, section "Non-Quickstart/State", is missing an import in the example.

```
# import Machine and State class
from transitions import State
```

It should be

```
# import Machine and State class
from transitions import Machine, State
```

PR #542, opened 2021-08-21
Missing "Machine" import

kauedg:
The main `README.md` file, section "Non-Quickstart/State", is missing an import in the example.

```
# import Machine and State class
from transitions import State
```

It should be

```
# import Machine and State class
from transitions import Machine, State
```

Fix for issue #541


aleneum:
Hello @kauedg,

thank you for taking the time to open the PR.

kauedg:
I'm the one who's thankful for this great module

Issue #549, opened 2021-09-29
Check may apply transition


artofhuman:
### Add method to check may model to be transited to target state.

Currently, we have a method to check the current state ex: `is_STOPPED()` but don't have a method to check if a model may be reached new state ex: `may_STOPPED()`

Example api:
``` python
machine.add_transition(
    "stop", "started",  "stopped",
)

timer.start()
timer.is_started() # => true

# new method
timer.may_stopped() # => true

timer.stop()

timer.may_stopped() # => false
```

**Additional context**
For example one of the popular ruby libs [AASM](https://github.com/aasm/aasm) has a similar API for check transitions

Example how we've implemented it in our code, but I think it should be in core lib to make client code more clear and useful 
```python
def may_to_state(self, state)
    self.state_machine.get_transitions(source=self.state.name, dest=state.name)
```

aleneum:
Hello again, @artofhuman 👋 !

> For example one of the popular ruby libs AASM has a similar API for check transitions

getting this functionality for transitions can be achieved by a) extending `_add_model_to_trigger` and optionally b) evaluating conditions to only return `True` when condition checks pass:

```python
from transitions import Machine, EventData


class MayMachine(Machine):

    def _can_trigger(self, model, *args, **kwargs):
        e = EventData(None, None, self, model, args, kwargs)

        return [trigger_name for trigger_name in self.get_triggers(model.state)
                if any(all(c.check(e) for c in t.conditions)
                       for t in self.events[trigger_name].transitions[model.state])]

    def _add_trigger_to_model(self, trigger, model):
        super()._add_trigger_to_model(trigger, model)
        self._checked_assignment(model, f"may_{trigger}", lambda: trigger in self._can_trigger(model))


m = MayMachine(states=["A", "B", "C"],
               transitions=[["go", "A", "B"],
                            dict(trigger='do', source='A', dest='C', conditions=lambda: False),
                            ["do", "B", "C"]],
               initial="A")

assert m.state == 'A'
assert m.may_go()
assert not m.may_do()
```

But you request this for destination states, right? This is not supported by AASM, is it?

aleneum:
>  I think overriding private methods is not good practice because they may be changed in future versions. 

fair enough 

Option A is something that is occasionally requested. Would that be sufficient for your use case?

artofhuman:
Yes 😄

aleneum:
`may` has been introduced in 0.9.0 and was recently pushed to master

PR #550, opened 2021-10-12
Generate methods may_<trigger> for check possible transition

artofhuman:
Ref #549

Hi, @aleneum. I create PR with feature `object.may_<trigger_name>` based on your sugeestion in the issue.


broglep-work:
We gave this PR a try and it does not seem to work with nested state machines

aleneum:
Hello @broglep-work,

thanks for the info. FYI: I will be back working on transitions after 15th of November.

artofhuman:
I've added this feature for nested states + more tests. I've never used `prepare` callbacks before as well. I'll try to check this too.

aleneum:
I merged your changes into the dev branch for 0.9 and added nested and async versions to the related machines. The tests pass so far. If you have some further test cases to integrated, let me know.

PR #566, opened 2022-03-04
typo

termim:


aleneum:
Hello @termim,

thank you for taking the time to open a PR for this. Everything that makes the documentation better is much appreciated!

PR #584, opened 2022-08-11
Create py.typed

### LINKED ISSUE #572: Add a py.typed file
Running mypy with:

```toml
[tool.mypy]
# ...
ignore_missing_imports = false
# ...
```

```
src/app/root_machine.py:6: error: Skipping analyzing "transitions": module is installed, but missing library stubs or py.typed marker  [import]
src/app/root_machine.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 3 source files)
ERROR: 1
```
On the surface the library appears to be well typed, would adding a `py.typed` file be ok?

spagh-eddie:
Fixes #572


aleneum:
Hello @spagh-eddie,

thank you for taking the time to create a PR for this. I have these files locally already but never managed to push them. There are still some inconsistencies considering typing but I hope it still will be useful in its current condition.

Cheers.

spagh-eddie:
> I hope it still will be useful in its current condition

so do I!

PR #623, opened 2023-07-02
chore: Update supported Python versions

ipeluffo:
This PR:
1. Removes Python 2.7 and 3.6 (and older) versions from the list of supported versions
2. Adds Python 3.11 version to the list of the supported versions
3. Does minimal changes to code to remove `six` and other required logic for the removed versions. I left other improvements like type hinting which are better supported in Python 3+ for a future PR

Note: [Python 3.7 has reached EOL](https://devguide.python.org/versions/) on 27/06/2023 but I left it in the list of supported versions since many libraries still support it.


aleneum:
Hi @ipeluffo,

sorry for not responding earlier and thank you for your PR. When you posted this, I checked the current version distribution reported by pypi:

Screenshot 2023-07-03 at 17 06 57

I guess it is time to drop 2.7 with only 2% users left. However, Python 3.6 still seems to be quite popular. 
Nevertheless dropping support for Python 3.5 and below in the next major release `0.10` makes sense to me.
I still have some bugfixes and feature requests left which I would like to tackle in an at least one 0.9.1 release first though. Adding 3.11 to the list of supported versions should happen ASAP of course.

aleneum:
Ooooh, they actually removed Python 2.7 from GHA actions. Well, some of the changes need to be done right away then I guess.

PR #653, opened 2024-04-14
remove Python 2 crumbs

a-detiste:


a-detiste:
@ipeluffo  @aleneum This is the continuation of https://github.com/pytransitions/transitions/pull/623

ipeluffo:
PR looks sensible but there are still some references to Python 2 in the code. Not sure if this would break CI or any tests

a-detiste:
It should be fine ... cleanup is an iterative process.

PR #673, opened 2024-05-28
fix `pytest_ignore_collect` not to block default pytest code

mgorny:
Fix `pytest_ignore_collect` hook implementation to return `None` rather than `False` when the path ought not to be ignored.  This is necessary to enable the default pytest implementation of `pytest_ignore_collect()` to be used.  Otherwise, the default rules are never applied and e.g. `--ignore` does not work.

I've also reported https://github.com/pytest-dev/pytest/issues/12383 about the misleading documentation.


mgorny:
Thanks!

Issue #507, opened 2021-02-01
Improve Parallel State Support


aleneum:
Right now, states must be direct descendants from the same state to be entered in parallel. I would be a significant improvement if this restriction could be lifted. For instance, transitions could pass lists of states (or state names) as `dest` to be entered in parallel. 

``` python
m = HierarchicalMachine(states=['A', 'B', 'C'], transitions=[['go', 'A', ['B', 'C']]], initial='A')
assert m.state == 'A'
m.go()
assert m.state = ['B', 'C']
```

thedrow:
Nested state machines cannot be parallel:
```
m1=HierarchicalMachine(states=['a', 'b'], initial='a')
>>> m2=HierarchicalMachine(states=['d', 'e'], initial='b')
>>> m2=HierarchicalMachine(states=['d', 'e'], initial='d')
>>> m3=HierarchicalMachine(states={'name': 'c', 'children': [m1, m2], 'parallel': [m1, m2]}, initial='c')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-q-gDbjwh-py3.9/lib/python3.9/site-packages/transitions/extensions/nesting.py", line 339, in __init__
    _super(HierarchicalMachine, self).__init__(*args, **kwargs)
  File "/home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-q-gDbjwh-py3.9/lib/python3.9/site-packages/transitions/core.py", line 565, in __init__
    self.add_states(states)
  File "/home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-q-gDbjwh-py3.9/lib/python3.9/site-packages/transitions/extensions/nesting.py", line 488, in add_states
    new_state.initial = [s if isinstance(s, string_types) else s['name'] for s in state_parallel]
  File "/home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-q-gDbjwh-py3.9/lib/python3.9/site-packages/transitions/extensions/nesting.py", line 488, in <listcomp>
    new_state.initial = [s if isinstance(s, string_types) else s['name'] for s in state_parallel]
TypeError: 'HierarchicalMachine' object is not subscriptable
```

aleneum:
you cannot pass more than one hsm to the children/states keyword since the states of the hsm are referenced in the named state. The reason is that there might be naming collisions of states defined in `m1` and `m2`.

```python
[...]

PR #687, opened 2024-08-23
fix: allow a list of machines in "parallel" 

aleneum:
… and thus initial to be a (list of) machines

- part of #507
- note that machine states must be unique!


aleneum:
Currently, the 'flat' nesting does not work with the scope resolution in place.

PR #696, opened 2025-05-01
Apply string_type checks and Enum checks in the correct order

### LINKED ISSUE #695: Using StrEnum for states with a `HierarchicalMachine` induces unexpected behaviors.
Hi there, first of all I'd like to take this occasion to thank you for your awesome work with this lib. We use statecharts extensively at my company,
and this lib is a lifesaver. Been using it for years!

**Describe the bug**

Using StrEnum for states with a `HierarchicalMachine` induce the following unexpected behaviors at the moment:

- Using the `initial`  parameter to set the initial state causes the code to crash when resolving the initial_state
- `get_transitions` does not include the expected transitions.
- The wrong Exception is raised when using a separator in a StrEnum Member Name

These behaviors seem to be caused by `string_types` checks being applied before `Enum` checks at several points.
I listed the checks I've identified so far in "Additional Context"

**Minimal working example**

```python

import enum
from transitions.extensions.nesting import HierarchicalMachine, NestedState

NestedState.separator = "__"


class States(enum.StrEnum):
    home = "a home"
    on_a_walk = "on a walk"
    walking = "currently walking"
    sitting = "currently sitting"
    currently__sitting = "currently sleeping"


class Event:
    go_home = "go_home"
    go_walk = "go_walk"
    sit = "sit"
    walk = "walk"


transitions = [
    {
        "source": States.on_a_walk,
        "dest": States.home,
        "trigger": Event.go_home,
        "before": [lambda: print("going home")],
    },
    {
        "source": States.home,
        "dest": States.on_a_walk,
        "trigger": Event.go_walk,
        "before": [lambda: print("going home")],
    },
    {"source": States.home, "dest": States.on_a_walk, "trigger": Event.go_walk},
    {
        "source": States.walking,
        "dest": States.sitting,
        "trigger": Event.sit,
        "before": [lambda: print("sitting")],
    },
    {"source": States.sitting, "dest": States.walking, "trigger": Event.walk},
]

# === Uncomment to observe crash at intialization
# machine_1 = HierarchicalMachine(
#     states=[
#         States.home,
#         {
#             "name": States.on_a_walk,
#             "children": [States.walking, States.sitting],
#             "initial": States.walking,
#         },
#     ],
#     initial=States.walking,
#     transitions=transitions,
# )


# This crashes with a KeyError("currently") instead of
# The ValueError explaining not to use the separator

# machine_2 = HierarchicalMachine(
#     states=[
#         States.currently__sleeping,
#         States.home,
#         {
#             "name": States.on_a_walk,
#             "children": [States.walking, States.sitting],
#             "initial": States.walking,
#         },
#     ],
#     transitions=transitions,
# )

machine_3 = HierarchicalMachine(
    states=[
        States.home,
        {
            "name": States.on_a_walk,
            "children": [States.walking, States.sitting],
            "initial": States.walking,
        },
    ],
    transitions=transitions,
)

# set_state works as expected
machine_3.set_state(States.walking)

# Should include a transitions for go_home and sit
# as well as to_* transitions
# assert machine_3.get_transitions("", States.walking, delegate=True) != []

# This is needed instead
target_transitions = machine_3.get_transitions(
    "", machine_3._get_enum_path(States.walking), delegate=True
)
```

**Expected behavior**

I expect `StrEnum` to behave as normal Enum states

**Additional context**

The root cause of these issues seems to be the order in which the
`isinstance(<state_related_variable>, string_types)` and `isinstance(<state_related_variable>, Enum)` are
ordered. At several places, `string_types` is checked first, and a member of a StrEnum is indeed as string types.

So far, I have identified this pattern at the following places:

- in \_\_call\_\_\*:

```python

    def __call__(self, to_scope=None):
        if isinstance(to_scope, string_types):
            state_name = to_scope.split(self.state_cls.separator)[0]
            state = self.states[state_name]
            to_scope = (state, state.states, state.events, self.prefix_path + [state_name])
        elif isinstance(to_scope, Enum):
            state = self.states[to_scope.name]
            to_scope = (state, state.states, state.events, self.prefix_path + [to_scope.name])
        elif to_scope is None:
            if self._stack:
                to_scope = self._stack[0]
            else:
                to_scope = (self, self.states, self.events, [])
        self._next_scope = to_scope

```

When setting up a scoping context, if `to_scope` is an Enum, it will go through the first, if, `to_scope.split(self.state_cls.separator)[0]` will return
the _value_ of the enum member instead of the name, and the state lookup will fail (hence the `KeyError`)

This happens during the recursive resolution of the initial_state. However, it does NOT happen when triggering an event, as build_state_tree use the state name,
which mean to_scope is always a string when resolving the trigger recursively in `_trigger_event_nested`

_In get_transitions_

```python

        with self():
            source_path = [] if source == "*" \
                else source.split(self.state_cls.separator) if isinstance(source, string_types) \
                else self._get_enum_path(source) if isinstance(source, Enum) \
                else self._get_state_path(source)
            dest_path = [] if dest == "*" \
                else dest.split(self.state_cls.separator) if isinstance(dest, string_types) \
                else self._get_enum_path(dest) if isinstance(dest, Enum) \
                else self._get_state_path(dest)
            matches = self.get_nested_transitions(trigger, source_path, dest_path)
            # only consider delegations when source_path contains a nested state (len > 1)
            if delegate is False or len(source_path) < 2:
                return matches
            source_path.pop()
            while source_path:
                matches.extend(self.get_nested_transitions(trigger,
                                                           src_path=source_path,
                                                           dest_path=dest_path))
                source_path.pop()
            return matches

```

In the same fashion, when source is an enum, the source_path will be `[<enum member value>]` instead of the result of \_get_enum_path because the `string_types` is matched first

_In add_states_

```python

        for state in listify(states):
            if isinstance(state, Enum):
                if isinstance(state.value, EnumMeta):
                    state = {'name': state, 'children': state.value}
                elif isinstance(state.value, dict):
                    state = dict(name=state, **state.value)
            if isinstance(state, string_types):
                self._add_string_state(state, on_enter, on_exit, ignore, remap, **kwargs)
            elif isinstance(state, Enum):
                self._add_enum_state(state, on_enter, on_exit, ignore, remap, **kwargs)
            elif isinstance(state, dict):
                self._add_dict_state(state, ignore, remap, **kwargs)
            elif isinstance(state, NestedState):
                if state.name in self.states:
                    raise ValueError("State {0} cannot be added since it already exists.".format(state.name))
                self.states[state.name] = state
                self._init_state(state)
            elif isinstance(state, HierarchicalMachine):
                self._add_machine_states(state, remap)
            elif isinstance(state, State) and not isinstance(state, NestedState):
                raise ValueError("A passed state object must derive from NestedState! "
                                 "A default State object is not sufficient")
            else:
                raise ValueError("Cannot add state of type {0}. ".format(type(state).__name__))

```

And finally, when adding StrEnum states, the `_add_string_state` is used, which produces the wrong error message when using a separator in the name. This is
the only consequences here as far as I can tell

I've started to work on a PR to switch the order of the checks, I'll add it shortly after publishing this issue. Happy to hear your thoughts :)

Thanks again for maintaining this great project! 🎉

jbrocher:
Proposal to fix #695. See issues for details about the problems this PR solves. 

I've added a test that inherit from the current Nested Enum test, and changes the set-up to use StrEnum. Running it produces a crash when resolving the initial state, and a fail test on `get_transitions`. The following commits fix these failures. 

One more change is needed to correctly use `_add_enum_state`, I'll try to add a test to assert that the correct exception is raised when using the separator in an Enum Member name first.

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import Enum, StrEnum
 from unittest import TestCase, skipIf
 
 from transitions.core import Machine
@@ -383,6 +383,19 @@ class TestNestedStateEnums(TestEnumsAsStates):
 
         ref_state = [P.P1, [Q.Q1, [[[X.X1, X.X2], B.B2], A.A2]]]
         self.assertEqual(ref_state, m.state)
+
+
+class TestNestedStateStrEnums(TestNestedStateEnums):
+
+    def setUp(self):
+        super().setUp()
+
+        class States(StrEnum):
+            RED = "red"
+            YELLOW = "yellow"
+            GREEN = "green"
+
+        self.States = States
 
 
 @skipIf(enum is None or (pgv is None and gv is None), "enum and (py)graphviz are not available")
--- a/transitions/extensions/nesting.py
+++ b/transitions/extensions/nesting.py
@@ -695,12 +695,12 @@ class HierarchicalMachine(Machine):
         """
         with self():
             source_path = [] if source == "*" \
-                else source.split(self.state_cls.separator) if isinstance(source, string_types) \
                 else self._get_enum_path(source) if isinstance(source, Enum) \
+                else source.split(self.state_cls.separator) if isinstance(source, string_types) \
                 else self._get_state_path(source)
             dest_path = [] if dest == "*" \
-                else dest.split(self.state_cls.separator) if isinstance(dest, string_types) \
                 else self._get_enum_path(dest) if isinstance(dest, Enum) \
+                else dest.split(self.state_cls.separator) if isinstance(dest, string_types) \
                 else self._get_state_path(dest)
             matches = self.get_nested_transitions(trigger, source_path, dest_path)
             # only consider delegations when source_path contains a nested state (len > 1)
--- a/transitions/extensions/nesting.py
+++ b/transitions/extensions/nesting.py
@@ -417,13 +417,13 @@ class HierarchicalMachine(Machine):
         )
 
     def __call__(self, to_scope=None):
-        if isinstance(to_scope, string_types):
+        if isinstance(to_scope, Enum):
+            state = self.states[to_scope.name]
+            to_scope = (state, state.states, state.events, self.prefix_path + [to_scope.name])
+        elif isinstance(to_scope, string_types):
             state_name = to_scope.split(self.state_cls.separator)[0]
             state = self.states[state_name]
             to_scope = (state, state.states, state.events, self.prefix_path + [state_name])
-        elif isinstance(to_scope, Enum):
-            state = self.states[to_scope.name]
-            to_scope = (state, state.states, state.events, self.prefix_path + [to_scope.name])
         elif to_scope is None:
             if self._stack:
                 to_scope = self._stack[0]


aleneum:
Hi @jbrocher, 

thank you for your report and PR. I think this won't solve #695 entirely since there are still issues with `StrEnum` with spaces but you are absolutely correct that we should check for Enums first since `StrEnum` is also a string type.
I will check what I can do about spaces in enum values.

This PR looks good to me. Thanks again for taking the time to document your findings and your contribution to this project.

PR #709, opened 2025-06-20
Fix #708: Mermaid graph code generation should be prepared for states whose names have spaces or dashes

### LINKED ISSUE #708: Mermaid graph code generation should be prepared for states whose names have spaces or dashes
Mermaid code does not accept states with spaces or dashes. The code should be prepared for this by, at least, sanitizing the names before generating the Mermaid code.

oEscal:
Fix for #708


aleneum:
Hmm.. the rebase/merge with the feature branch did not work, unfortunately. I don't think a '_sanitize_name' (right now, this is just a placeholder, isn't it) function is required everywhere unless they are some characters that are also forbidden in labels. I will check it out.

aleneum:
Closing this since the feature has been committed to the feature branch.

Issue #713, opened 2025-07-03
Community: Should AsyncMachine continue handling `CancelledErrors` raised by unfinished events silently?


aleneum:
### Current Behavior
As of now `AsyncMachine` will handle `CancelledErrors` caused by cancelling tasks that are still running while a new event is triggered.

### Question
**Do you consider this useful and convenient or would you prefer to handle cancellation events yourself via for instance try/except or `on_exception`?**

### Example
```python
from transitions.extensions import AsyncMachine
import asyncio


class Model:

    def __init__(self):
        self.data = []

    async def doing_task_1(self):
        print("Doing task 1...")
        await asyncio.sleep(2)
        self.data.append(1)
        await self.done()

    async def doing_task_2(self):
        print("Doing task 2...")
        await asyncio.sleep(2)
        self.data.append(2)
        await self.done()

    async def doing_task_3(self):
        print("Doing task 3...")
        self.data.append(3)
        await self.done()

states = ["idle", "busy"] 
transitions = [
    {"trigger": "do_task_1", "source": "*", "dest": "busy", "after": "doing_task_1"},
    {"trigger": "do_task_2", "source": "*", "dest": "busy", "after": "doing_task_2"},
    {"trigger": "do_task_3", "source": "*", "dest": "busy", "after": "doing_task_3"},
    {"trigger": "done", "source": "busy", "dest": "idle"},
]

model = Model()
m = AsyncMachine(model=model, states=states, transitions=transitions, initial="idle")


async def main():
    asyncio.create_task(model.do_task_1())  # won't finish
    await asyncio.sleep(0.1)
    asyncio.create_task(model.do_task_2())  # won't finish
    await asyncio.sleep(0.1)
    asyncio.create_task(model.do_task_3())  # does finish
    await asyncio.sleep(0.1)
    assert model.is_idle()
    assert model.data == [3]

asyncio.run(main())
```

Salier13:
As the source of this discussion, I believe that handling the exeption ourselves would be better.
In a state machine context, having a transition from A to B be halfway done and then cancelling it to start a transition from A to C results in nether B or C unless both are aware that it is a possibility that their starting state may be a dirty A and that their transition may get cancelled.

```python
import asyncio
from transitions.extensions.asyncio import AsyncMachine


class Model:

    def __init__(self):
        self.counter = 0

    async def on_enter_a(self):
        self.counter = 0

    async def on_exit_a(self):
        self.counter += 1
        await asyncio.sleep(1)

[...]

Issue #704, opened 2025-06-13
Async TimeoutError gets caught inappropriately


Salier13:
When using the AsyncMachine, any asyncio.TimeoutError that gets raised during a manipulation of the state machine gets eaten by it and process outside of state machine keep going as if nothing happened.

```python
import asyncio
from transitions.extensions.asyncio import AsyncMachine

class Model:
    async def on_enter_on(self):
        await asyncio.sleep(1)
        assert False, "Shouldn't be reachable due to timeout"

machine = AsyncMachine(
    Model(),
    states=["off", "on"],
    transitions=[{"trigger": "start", "source": "off", "dest": "on"}],
    initial="off"
)

async def foo():
    await machine.model.start()
    assert False, "Shouldn't be reachable due to timeout" # <-- This assert trigger

async def main():
    try:
        await asyncio.wait_for(foo(), timeout=0.5)
    except asyncio.TimeoutError:
        return         # expected case
    assert False, "Expected a TimeoutError"

asyncio.run(main())
```

I expect the Timeout exception to be raised if no particular comportment for exceptions is indicated. Particularly in the case where the timeout extension is not used or is not specified for the current transition.

I did not test this with https://github.com/pytransitions/transitions/pull/686 so unsure if it is fixed with that timeout rework. But if it is could a release be planned soon?

aleneum:
The example above works now as you expect. Only cancellation events with a msg set to `asyncio.CANCELLED_MSG` will be caught by transitions. This fix only works for `Python 3.11` and above where `Task.cancel` has a `msg` parameter.

aleneum:
Sorry for the noise in your inbox. I decided to make this a new issue (#713). Feel free to comment here or there. Closing this again.

Issue #619, opened 2023-04-07
"on_exception" is not invoked for "KeyboardInterrupt"


aleneum:
Catch `BaseException` instead of `Exception` during state transition. 
Even though user user exceptions should be derived from `Exception` [[1](https://docs.python.org/3/library/exceptions.html#built-in-exceptions)], some exceptions such as `KeyboardInterrupt` are not. I do not see an imminent problem with doing this switch.

Discussed in https://github.com/pytransitions/transitions/discussions/618

Issue #259, opened 2017-10-03
Asyncio support


lromor:
Hello,
are there any news on the asyncio support? 

thank you,
-l

aleneum:
Hello @lromor,

unfortunately making `transitions` run async callbacks requires some major refactoring. Otherwise the async version would more or less copy big chunks of code from the core machine which is a lot harder to maintain.

However, we planned to do this for a while. To check if we are on the right track I got some questions for you:

* What do you need to call asynchronously? The callbacks in `prepare`,`conditions`,`before`,`after` or the transitions themselves?
 * Who is going to manage the event loop? If most code runs synchronously and only async callbacks have to be synchronised somehow, `Machine` could handle the event loop.

lromor:
Hello @aleneum , 
first of all, I forgot to thank you for your work, it's a very well done project.
I will be a bit verbose on this post, sorry for that.

So, we are working on a real machine and we wanted to have a more consistent way to define its behavior.

# Introduction to our problem 
(it's just to share our view, it may be important for you to understand if we may match or not transitions goals):

One of the biggest problems introduced by the async (or also with safe multithread) is that for big projects (without an FSM) a lot of states variables are spread around the program and because of that it becomes an extremely difficult task pin down perfectly the machine behavior. Similar problems are encountered for example while building async/event driven web interfaces. It can really become a mess when you start triggering a lot of changes at the same time and you don't really know what is the flow of your program anymore (see flux paradigm).
So we decided to introduce an FSM that centralizes all of this logic making the program hopefully more consistent and ease the testing process.

Regarding your questions:

1.  **Short answer:** Before/on_enter after/on_exit is the way to go.
**Long answer**: I think that for this kind of application, **actions should be [states](https://en.wikipedia.org/wiki/Finite-state_machine#/media/File:Fsm_Moore_model_door_control.svg)**. 
We require the possibility to change an action if for example during "running" a stop event occurs. Transitions should always be immediate. 
I think that modeling actions with transitions may introduce some flaws, that would lead us to merge the states logic to the transitions one (ie we will need to define if allow or do not allow to transition from a transition to another state).
I think also that in general, in order to avoid to keep tracking all the asynchronous tasks spawned by a state a single a better approach of the callbacks on_enter/on_exit  would be to allow an optional state's task. I can send you a snippet of what I mean. 

2.  **Short answer:** I think that the async loop should be managed externally. Mostly because there's no real reason to force this requirement.

aleneum:
[...]

PR #731, opened 2025-09-08
Swallowing a BaseException is not allowed.

smurfix:
Python basically forbids swallowing a `BaseException` that's not an `Exception`.

(You do not want a `KeyboardInterrupt` or a `SystemExit` to not terminate the program.)

Also, using asyncio, you must always propagate a cancellation.

### DIFF
--- a/transitions/core.py
+++ b/transitions/core.py
@@ -422,7 +422,7 @@ class Event(object):
             event_data.error = err
             if self.machine.on_exception:
                 self.machine.callbacks(self.machine.on_exception, event_data)
-            else:
+            if not self.machine.on_exception or not isinstance(err, Exception):
                 raise
         finally:
             try:
@@ -917,7 +917,7 @@ class Machine(object):
                     event_data.error = err
                     if self.on_exception:
                         self.callbacks(self.on_exception, event_data)
-                    else:
+                    if not self.on_exception or not isinstance(err, BaseException):
                         raise
         return False
 
--- a/transitions/extensions/asyncio.py
+++ b/transitions/extensions/asyncio.py
@@ -23,6 +23,7 @@ import warnings
 from collections import deque
 from functools import partial, reduce
 import copy
+from concurrent.futures import CancelledError
 
 from ..core import State, Condition, Transition, EventData, listify
 from ..core import Event, MachineError, Machine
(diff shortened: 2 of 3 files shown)


aleneum:
Hello @smurfix,

> Python basically forbids swallowing a BaseException that's not an Exception.

is there a PEP or something where I can look this up? 
I just ask because some people specfically asked for catching `BaseException` (e.g. #619). 
Just so that I have a solid argument when somebody asks why this has been changed.

smurfix:
There doesn't seem a PEP-or-whatever documenting this, unfortunately.

https://docs.python.org/3/tutorial/errors.html states

> Exceptions which are not subclasses of [Exception](https://docs.python.org/3/library/exceptions.html#Exception) are not typically handled, because they are used to indicate that the program should terminate. […]

There are a couple of `BaseException` instances (e.g. GeneratorExit or StopIteration) which don't actually exit the whole program, but they do exit a particular scope and thus catching them and re-raising them somewhere else doesn't make much sense. Same for async cancellation. In that vein I think I'll submit a patch that elaborates on this.

I'm not saying not to call the exception handler, which is what #619 wants us to. My patch specifically *does* call the handler. It just always propagates the exception instead of eating it, after doing so.

smurfix:
I have started https://discuss.python.org/t/forbid-forwarding-or-not-re-raising-a-baseexception/103646 .

aleneum:
Hi @smurfix,

> I'm not saying not to call the exception handler, which is what https://github.com/pytransitions/transitions/issues/619 wants us to. My patch specifically does call the handler. It just always propagates the exception instead of eating it, after doing so.

got it. I absolutely see why 'handling' all exceptions causes unwanted behaviour (`KeyboardInterrupt` not stopping execution is very unconvenient imo) but I had the impression that's what users want since transitions used to handle `Exceptions` only in the past.
 
> There are a couple of BaseException instances (e.g. GeneratorExit or StopIteration) which don't actually exit the whole program, but they do exit a particular scope and thus catching them and re-raising them somewhere else doesn't make much sense. Same for async cancellation. In that vein I think I'll submit a patch that elaborates on this.

`asyncio.CancelledError` is a bit more tricky since it's part of the intended workflow that transitions may cancel long-running callbacks (discussed in #259 as far as I remember). But you are not the only one confused by this (#704, #713). I opted for 'marking' cancellations caused by transitions and only raise cancellations caused by other factors. I happily remove complex exception handling behavior as long as this doesn't render some users' workflow impossible.

> I have started https://discuss.python.org/t/forbid-forwarding-or-not-re-raising-a-baseexception/103646 .

looking forward to feedback

